PR #21, opened 2015-01-22
Fix missing state when starting a run

smathieu:


smathieu:
Had to disable some tests that did not (and never) worked.

PR #23, opened 2015-01-29
Give some much needed love to rainforest-cli

smathieu:
This includes:

- A complete refactoring of the code
- Set a user agent for all requests
- Use the "is_final_state" API option to know when to stop polling
- Display a link to the UI for detail results

@jipiboily


spec/git_trigger_spec.rb
```diff
@@ -1,13 +1,25 @@
...
+      setup_test_repo
+      begin
+        spec.call
+      ensure
+        Dir.chdir(default_dir)
+      end
+    end
 
-      Dir.chdir(File.join([default_dir, 'spec', 'test-repo']))
+    # Commented because it kills CI
```
jipiboily: Why? Can you add details or a link to the failure plz?

jipiboily:
Looks good to me. Do you feel confident that our test suite in this would catch errors of this refactor? It feels like it.

PR #24, opened 2015-03-06
Fix bad API_URL constant

shosti:
@rainforestapp/devs


ukd1:
@shosti can you bump the version for the gem pls

PR #32, opened 2015-09-11
Add smart folders

nxvl:


maciejgryka:
@nxvl Did you try running a smart folder from our suite?

nxvl:
Not yet.

lib/rainforest/cli/runner.rb
```diff
@@ -92,6 +92,8 @@ def make_create_run_options
           # Not using git_trigger, so look for the
           if !options.tags.empty?
             post_opts[:tags] = options.tags
+          elsif !options.folder.nil?
+            post_opts[:smart_folder_id] = @options.folder.to_i
```
maciejgryka: So the back-end already supports this and we only need to send more post data?

nxvl: yup, that's how we start runs on smart folders from the app.

nxvl:
@maciejgryka started a run succesfully!

epaulet:
:shipit:

PR #51, opened 2016-02-17
Create all parent folders required for test folder

### LINKED ISSUE #50: New test will error if spec folder does not exist
It should just create the default folder if it doesn't already exist (it already creates the subfolder for you).

It will make it easier for a newer user to be successful without running into too many error because they forgot to specify a folder.

epaulet:
And some other code clean up

fixes #50

### DIFF
--- a/lib/rainforest/cli/options.rb
+++ b/lib/rainforest/cli/options.rb
@@ -1,13 +1,7 @@
+# frozen_string_literal: true
 require 'optparse'
 
 module RainforestCli
-  class BrowserException < Exception
-    def initialize browsers
-      invalid_browsers = browsers - OptionParser::VALID_BROWSERS
-      super "#{invalid_browsers.join(', ')} is not valid. Valid browsers: #{OptionParser::VALID_BROWSERS.join(', ')}"
-    end
-  end
-
   class OptionParser
     attr_writer :file_name, :tags
     attr_reader :command, :token, :tags, :conflict, :browsers, :site_id, :environment_id,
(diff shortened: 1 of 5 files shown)


lib/rainforest/cli/options.rb
```diff
@@ -133,9 +121,21 @@ def foreground?
...
+          raise ValidationError, 'You must pass your API token using: --token TOKEN'
+        end
+      end
+
+      if browsers
+        raise BrowserException, browsers unless (browsers - VALID_BROWSERS).empty?
+      end
+
+      if folder
+        # Allow it to start with `./` but not with just `/`
```
shosti: Why?

shosti:
looks good other than question...

lib/rainforest/cli/options.rb
epaulet:
```
[2] pry(main)> Dir.exist?('app')
=> true
[3] pry(main)> Dir.exist?('/app')
=> false
[4] pry(main)> Dir.exist?('./app')
=> true
```

shosti: Yeah, obviously, but people might also want to use absolute paths
(e.g. /home/deploy/myapp/app).

epaulet: I figured this would only be used when at the root of the project that holds your Rainforest tests... but that's a good point.

epaulet:
@shosti Better now?

jbarber:
@epaulet I just ran into this problem and was about to submit my own PR, before I saw yours! :)

FYI there is a `FileUtils.mkdir_p()` [method](http://ruby-doc.org/stdlib-2.3.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-mkdir_p) which I think provides the same functionality as `create_test_folder`.

epaulet:
Good call.

epaulet:
@shosti ping!

PR #56, opened 2016-02-20
Spruce up Validation

### LINKED ISSUE #54: Validate does not support embedded test ids

epaulet:
fixes #54


.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: ruby
 rvm:
   - 2.3.0
-  - 2.2.3
```
ukd1: why remove - some customers may / will have older versions of Ruby

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Rainforest CLI Changelog
 
+## 1.2.2 - WIP
```
ukd1: No WIP in Changelog!

.rubocop.yml
```diff
@@ -1,2 +1,5 @@
 inherit_gem:
   rf-stylez: ruby/rubocop.yml
+
+Style/FileName:
+  Exclude: [rainforest-cli.gemspec]
```
shosti: :question: what's the beef?

.ruby-version
```diff
@@ -1 +0,0 @@
-2.3.0
```
shosti: Why?

epaulet: Since we are supporting all the way down to 2.0, I didn't want to specify 2.3 for development.

lib/rainforest/cli/test_files.rb
```diff
@@ -6,13 +6,16 @@ class RainforestCli::TestFiles
   attr_reader :test_folder, :test_paths, :test_data
 
   def initialize(test_folder = nil)
-    @test_folder = test_folder || DEFAULT_TEST_FOLDER
+    test_folder ||= DEFAULT_TEST_FOLDER
+    # remove trailing slash
+    @test_folder = test_folder.gsub(/#{File::SEPARATOR}$/, '')
```
shosti: This is a bit confusing (at first glance, it looks like it removes all separators even though it doesn't). I think `File.expand_path` should do what you want, right? (Also avoiding relative path weirdness.)

rainforest-cli.gemspec
```diff
@@ -7,8 +7,8 @@ require 'rainforest/cli/version'
 Gem::Specification.new do |spec|
   spec.name          = 'rainforest-cli'
   spec.version       = RainforestCli::VERSION
-  spec.authors       = ['Simon Mathieu', 'Russell Smith']
-  spec.email         = ['simon@rainforestqa.com', 'russ@rainforestqa.com']
+  spec.authors       = ['Russell Smith', 'Edward Paulet']
```
shosti: :+1: lol

.rubocop.yml
epaulet: The filename is dashed because the gem name is dashed.

.ruby-version
shosti: I think we should keep it for development (to avoid odd incompatibilities) and possibly add explicit Ruby version tests on travis?

.rubocop.yml
shosti: OK makes sense.

shosti:
I'm a little confused as to what the actual bug was and how this fixes it. Mind giving an overview?

epaulet:
Yeah. I ended up doing a lot of housekeeping in this repo just by virtue of having to touch a lot of the code.

The problem was that if you embedded a test whose rfml id isn't in the repo, it would be an infinite loop inside of `RainforestCli::TestImporter#upload`.

```
-    upload_groups = []
 -    unordered_tests = []
 -    queued_tests = test_files.test_data.dup
 -
 -    until queued_tests.empty?
 -      new_ordered_group = []
 -      ordered_ids = upload_groups.flatten.map(&:rfml_id)
 -
 -      queued_tests.each do |rfml_test|
 -        if (rfml_test.embedded_ids - ordered_ids).empty?
 -          new_ordered_group << rfml_test
 -        else
 -          unordered_tests << rfml_test
 -        end
 -      end
 -
 -      upload_groups << new_ordered_group
 -      queued_tests = unordered_tests
 -      unordered_tests = []
 -    end
```

This fixes it by adding a check to make sure that each loop is actually making a difference. If not, it raises an error and tells you which files are embedding non-existent tests.

epaulet:
@shosti does that make sense?

shosti:
It does make sense, it just seems awfully complicated... (Looking it over again.)

lib/rainforest/cli/test_importer.rb
```diff
@@ -312,4 +292,49 @@ def create_test_obj(rfml_test)
...
+      queued_tests.each do |rfml_test|
+        if (rfml_test.embedded_ids - ordered_ids).empty?
+          new_ordered_group << rfml_test
+        else
+          unordered_tests << rfml_test
+        end
+      end
+
+      # If all the queued tests make it to the unordered tests group, then
+      # they contain non-existent RFML ids.
```
shosti: This isn't obvious to me at all. Can you either explain it better of figure out a more obvious test?

In general, I feel like this whole method would be much clearer if it were recursive--I can't really keep track of what all the different groups are, and they seem to be kind of oddly named. What does this method do exactly?

epaulet: I'll try refactoring to a clearer approach.

epaulet:
I extracted the upload process to it's own class and tried to name things a bit better. Recursion would be very tricky to do here if you want to be able to upload as many tests as possible with Parallel.

lib/rainforest/cli/uploader.rb
```diff
@@ -0,0 +1,162 @@
...
+  def validate_embedded_tests!
+    contains_nonexistent_ids = rfml_tests.select { |t| (t.embedded_ids - all_rfml_ids).any? }
+
+    if contains_nonexistent_ids.any?
+      raise TestsNotFound.new(contains_nonexistent_ids.map(&:file_name))
+    end
+  end
+
+  # Prioritize embedded tests before other tests
+  def make_upload_groups
```
shosti: I still just don't quite understand what this is doing. Is it depth first search for embedded tests?

lib/rainforest/cli/uploader.rb
```diff
@@ -0,0 +1,162 @@
...
+      upload_group_in_parallel(upload_groups.first)
+    else
+      upload_groups_sequentially(upload_groups)
+    end
+  end
+
+  private
+
+  def validate_embedded_tests!
+    contains_nonexistent_ids = rfml_tests.select { |t| (t.embedded_ids - all_rfml_ids).any? }
```
shosti: This is much clearer, thanks.

lib/rainforest/cli/uploader.rb
epaulet: Then it's probably poorly written. Upload order matters because if you embed a parent before a child exists in the database, it'll fail. This is to set up the test priority.

Let me try breaking the loop out into its own test to see if that's more clear.

epaulet:
@shosti is that more clear?

epaulet:
@shosti Does this approach make more sense? The project directory could use some more splitting up, but i think it's out of scope for this PR.

epaulet:
Also, I'm going to extract validation to its own class in the next PR. I just don't want to pack this PR with anything else.

epaulet:
Validating when folder does not exist:
```
$ bin/rainforest validate
I, [2016-03-08T10:13:14.827879 #1501]  INFO -- : No test folder supplied. Using default folder: ./spec/rainforest
E, [2016-03-08T10:13:14.827988 #1501] ERROR -- : No tests found in directory: /Users/Edward/Desktop/RainforestQA/rainforest-cli/spec/rainforest
```

epaulet:
### Uploading

Uploading when a test doesn't exist:
```
$ bin/rainforest upload --token c743c7413178b4b78b8a0d08b5c71ae1
I, [2016-03-08T10:25:55.735982 #1806]  INFO -- : No test folder supplied. Using default folder: ./spec/rainforest
E, [2016-03-08T10:25:55.736086 #1806] ERROR -- : No tests found in directory: /Users/Edward/Desktop/RainforestQA/rainforest-cli/spec/rainforest
```

epaulet:
@ukd1 Can I have a review please?

PR #62, opened 2016-03-14
Remove 'ro' tag in favor of source 'rainforest-cli'

### LINKED ISSUE #61: Remove "RO" tag
The test model now supports an attribute called `source` and the appropriate source for RFML tests is `rainforest-cli`. We can use that source to designate whether a test should be allowed to be modified in the web application instead of the `ro` tag.

epaulet:
fixes #61

Do not merge until backend change is shipped first.

Edit: backend shipped.


ukd1:
lgtm but zero tests?

epaulet:
Updated specs for changing. Merging!

PR #67, opened 2016-03-21
Add support for ruby 1.9.3

### LINKED ISSUE #66: Support Ruby 1.9.3
CircleCI uses 1.9.3 as default: https://circleci.com/docs/language-ruby-on-rails
`The default version of Ruby is either 1.9.3-p448 or 1.8.7-p358, whichever we think is best for your project.`

epaulet:
Fixes #66

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,7 @@ rvm:
   - 2.0.0
   - 1.9.3
 
+before_install:
+  - gem update bundler
+
 script: bundle exec rspec spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ rvm:
   - 2.2.3
   - 2.1.8
   - 2.0.0
+  - 1.9.3
 
 script: bundle exec rspec spec
--- a/lib/rainforest/cli/http_client.rb
+++ b/lib/rainforest/cli/http_client.rb
@@ -33,7 +33,7 @@ module RainforestCli
       JSON.parse(response.body)
     end
 
-    def get(url, body = {}, options = { retries_on_failures: false })
+    def get(url, body = {}, options = {})
       wrap_exceptions(options[:retries_on_failures]) do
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -4,5 +4,9 @@ rvm:
   - 2.2.3
   - 2.1.8
   - 2.0.0
+  - 1.9.3
+
+before_install:
+  - gem update bundler
```
epaulet: Travis's bundler isn't updated so we were getting errors with several gems.

lib/rainforest/cli/http_client.rb
```diff
@@ -33,8 +33,8 @@ def post(url, body = {})
       JSON.parse(response.body)
     end
 
-    def get(url, body = {}, retries_on_failures: false)
-      wrap_exceptions(retries_on_failures) do
+    def get(url, body = {}, options = { retries_on_failures: false })
```
ziahamza: Any reason to move into an object just for one key?

ziahamza:
had to look what byebug was .. but other then that LGTM :)

lib/rainforest/cli/http_client.rb
shosti: I'd actually make `options` default to `{}` ("default options hash" has different semantics from "default arguments" and can be pretty confusing).

epaulet: It already was a hash to begin with. Just an implied hash <3 ruby

The original reason for it was because when an argument that is just True/False, it's not clear for the user of the method what the purpose is of that value is. At least to me.

epaulet: sure, i can make it an empty hash

shosti: Common pattern is something like:

```ruby
def foo(opts = {})
  opts = { opt1: :default1 }.merge(opts)
  ...
end
```

PR #72, opened 2016-03-30
Alow exporting individual tests

ziahamza:


ziahamza:
@epaulet Can I haz initially review on this?

This will basically allow exporting individual tests with the test id. i.e doing:
```
rainforest export --token ... 1234 2321 3212
```
where 1234, 2321 will be test IDs

lib/rainforest/cli/test_importer.rb
```diff
@@ -38,17 +38,22 @@ def threads
...
   def export
-    tests = Rainforest::Test.all(page_size: 1000)
-    p = ProgressBar.create(title: 'Rows', total: tests.count, format: '%a %B %p%% %t')
-    Parallel.each(tests, in_threads: threads, finish: lambda { |_item, _i, _result| p.increment }) do |test|
+    test_ids = []
+    if @options.tests.length > 0
+      test_ids = @options.tests
+    else
+      test_ids = Rainforest::Test.all(page_size: 1000).map { |t| t.id }
+    end
```
epaulet: Let's do this the Ruby way :sunglasses:

You can set test_ids directly from the return value of your `if` block (so you don't have to set it to an empty array first and manually set it in the if block). Or you can do a ternary if it's short enough.

```ruby
foo = if false
  baz
else
  lol
end

foo == lol #true
```

epaulet:
Please add specs for this. :)

epaulet:
BTW, this has conflicts because I changed the name of the file you changed.

epaulet:
@ziahamza Seems like we should be using RFML ID though right? The test's primary key is not specified in the RFML files.

ziahamza:
@epaulet No, this PR was there so that you can export an RFML test from an existing Rainforest test. So basically a test which has not been exported before. The user will only know the Test id as RFML id is presumably not added/shown in UI if its not exported before.

epaulet:
Ah, I was thinking upload. That makes sense. I'll re-review once we get new specs + all passing specs :)

ziahamza:
fixed/added basic specs. mind reviewing @epaulet ?

epaulet:
LGTM as long as you tested it on your own machine.

PR #73, opened 2016-03-30
get token from environment variable

valeriangalliat:
Rainforest token, if not given with `--token`, is fetched from the `RAINFOREST_API_TOKEN` environment variable.


epaulet:
:+1: 

Please add a quick spec and fix merge conflicts and this should be good to go :)

valeriangalliat:
Thnaks @epaulet, I updated the code with a spec.

epaulet:
Looks like our circlemator gem is breaking builds from contributors without write access to this repo so I removed that gem in the master branch. Could you merge master into this branch? That way it should be green.

valeriangalliat:
Done, test passed o/

PR #77, opened 2016-04-04
Redirect Option

### LINKED ISSUE #75: Support for browsers and redirect attributes on test
Right now we just set redirect to true on upload.

For uploads, browsers is ignored. But we don't have a static list of browsers to make it easy to choose from.

epaulet:
partially fixes #75


epaulet:
Putting this on hold until we figure out where we want the flag to appear in the RFML test.

epaulet:
Spec was agreed on by myself and @ukd1. Merging when green to get release out the door.

PR #79, opened 2016-04-05
Hotfix: add rainforest-cli attribute for empty tests

epaulet:
The flag makes sure the test is read-only, which is important in case something fails later along in the upload process.


ziahamza:
LGTM :+1:

PR #89, opened 2016-04-10
Update readme. Add --test-folder option usage.

marciaga:


epaulet:
Thanks for the contribution @marciaga!

Since that option is also usable for the `export` and `upload` commands, could you just add those into the description as well? Other than that 👍

marciaga:
@epaulet Updated, thanks!

Issue #80, opened 2016-04-06
Validate duplicate UUIDs.


epaulet:
Make sure that tests do not get overwritten.

PR #95, opened 2016-04-18
Adds dup rfml id validation

curtis-rainforestqa:
Addresses #80 by failing validation if there are tests with the same rfml_id.


epaulet:
:ship:

PR #106, opened 2016-05-19
App uploader

### LINKED ISSUE #101: Uploader for device-farm apps
If the ``app-source-url`` is a file, not a url, upload it to AWS S3 under Rainforest's account. We should probably add a API so we can get a pre-signed URL for that particular customer. Also, enable server side encryption.

marianafranco:
Fixes #101

Depends on https://github.com/rainforestapp/Rainforest/pull/8148

@rainforestapp/core @ukd1

### DIFF
--- a/lib/rainforest/cli/runner.rb
+++ b/lib/rainforest/cli/runner.rb
@@ -14,9 +14,7 @@ module RainforestCli
         CSVImporter.new(options.import_name, options.import_file_name, options.token).import
       end
 
-      app_source_url = options.app_source_url ? upload_app(options.app_source_url) : options.app_source_url
-
-      post_opts = make_create_run_options(app_source_url)
+      post_opts = make_create_run_options
 
       logger.debug "POST options: #{post_opts.inspect}"
       logger.info 'Issuing run'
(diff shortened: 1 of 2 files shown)


lib/rainforest/cli/runner.rb
```diff
@@ -14,7 +14,9 @@ def run
         CSVImporter.new(options.import_name, options.import_file_name, options.token).import
       end
 
-      post_opts = make_create_run_options
+      app_source_url = options.app_source_url ? upload_app(options.app_source_url) : options.app_source_url
```
epaulet: Seems like this would logically fit better under `make_create_run_options`. That way we can remove the new argument too.

lib/rainforest/cli/runner.rb
```diff
@@ -153,5 +155,55 @@ def get_environment_id url
 
       return environment['id']
     end
+
+    def upload_app(app_source_url)
```
epaulet: I think it would be a good idea to add specs for each of the logical branches in this method to make sure nothing breaks in a refactor.

marianafranco:
@epaulet please, include this in the next release ;)

PR #108, opened 2016-05-20
[DO NOT MERGE] Recommend site id instead of forcing for custom url

### LINKED ISSUE #103: Don't force site-id for custom URLs
A warning should be sufficient. There are legitimate use cases where you don't want to run just one site even though you are using a custom URL.

epaulet:
fixes #103


epaulet:
This seems like an edge use case... going to close this for now...

Issue #115, opened 2016-06-03
Exporting tests with --embed-tests doesn't keep the redirect flag


prayogoa:
https://app.intercom.io/a/apps/mfdgv5am/inbox/all/conversations/4945108314

epaulet:
Should be fixed as of #133

PR #117, opened 2016-06-07
Exporting tests with --embed-tests now keeps the redirect flag

theonegri:
Fixing issue https://github.com/rainforestapp/rainforest-cli/issues/115

@rainforestapp/customer-product 

I'm not confident with the spec I wrote tho. Please somebody make sure its good enough.

### DIFF
--- a/lib/rainforest/cli/exporter.rb
+++ b/lib/rainforest/cli/exporter.rb
@@ -55,6 +55,7 @@ class RainforestCli::Exporter
     when 'test'
       if @options.embed_tests
         file.puts '' unless index == 0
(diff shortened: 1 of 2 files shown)


epaulet:
The spec is fine for what you added. But now that I'm looking at this, I think it makes more sense to only include the redirection in some places (basically the same places that they would appear in the main app). Otherwise it might be confusing to see redirects in places that they aren't an option for in the dashboard. The rules for redirect flag locations are:

- Before the first real step of a test if the first step is actually an embedded test.
- Before an embedded test if it is not the first step of a test.

epaulet:
Going to merge this and we can continue making this nicer in another PR.

PR #124, opened 2016-06-08
Use faster rfml id endpoint

epaulet:


curtis-rainforestqa:
looks good, i'm a 👍 on green

PR #133, opened 2016-06-09
Add redirect flag for first steps after embedded tests

epaulet:
fixes #117


usami:
lgtm. I'm pretty sure the logic would work as expected.

Issue #129, opened 2016-06-09
validate isn't documented


juliankrispel:
There is 0 documentation for `rainforest validate`

epaulet:
Add documentation in #134

PR #134, opened 2016-06-10
Do not try to make API calls if token is not set for rfml lookups

### LINKED ISSUE #131: rainforest validate doesn't check if there is a token
When you execute `rainforest validate` without a token, HttpParty will through an error because there's a nil value in the headers:

```bash
/Users/juliiankrispel/.rbenv/versions/2.3.0/lib/ruby/2.3.0/net/http/header.rb:18:in `block in initialize_http_header': undefined method `strip' for nil:NilClass (NoMethodError)
```

epaulet:
fixes #131

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,17 @@ You can customize the amount of threads to use when making HTTP requests by sett
 API errors when fetching or updating multiple tests, lowering this value can help.
 The default value is `16`.
 
-Run all of your tests
+#### Commands
+
+###### Running Tests
+
+Run all tests.
 
 ```bash
 rainforest run all
 ```
 
-Run all in the foreground and report
+Run all in the foreground and report.
 
 ```bash
 rainforest run all --fg
@@ -49,6 +53,8 @@ Run all tests with tag 'run-me' and abort previous in-progress runs.
 ```bash
 rainforest run --tag run-me --fg --conflict abort
 ```
+
+###### Creating and Managing Tests
 
 Create new Rainforest test in RFML format (Rainforest Markup Language).
 
@@ -62,6 +68,14 @@ You may also specify a custom test title or file name.
 rainforest new "My Awesome Title"
 ```
 
+Validate your tests for syntax and correct RFML ids for embedded tests.
+Use the `--token` options or `RAINFOREST_API_TOKEN` environment variable
+to validate your tests against server data as well.
+
+```bash
+rainforest validate
+```
+
 Upload tests to Rainforest
 
 ```bash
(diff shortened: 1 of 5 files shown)


epaulet:
fixes #129 as well

lib/rainforest/cli/remote_tests.rb
```diff
@@ -13,29 +13,6 @@ def rfml_ids
     @rfml_ids ||= primary_key_dictionary.keys
   end
 
-  def tests
```
epaulet: this method doesn't seem to be used anymore. The endpoint is too slow anyway.

epaulet:
Old behavior is back!
```
$ bin/rainforest validate
I, [2016-06-10T00:20:25.672454 #48419]  INFO -- : No test folder supplied. Using default folder: ./spec/rainforest
I, [2016-06-10T00:20:25.678040 #48419]  INFO -- : Validating parsing errors...
I, [2016-06-10T00:20:25.678097 #48419]  INFO -- : Validating embedded test IDs...
I, [2016-06-10T00:20:25.678128 #48419]  INFO -- : No API Token set. Using local tests only...
I, [2016-06-10T00:20:25.679177 #48419]  INFO -- :
I, [2016-06-10T00:20:25.679201 #48419]  INFO -- : [VALID]
```

spec/exporter_spec.rb
```diff
@@ -3,7 +3,7 @@
   let(:options) do
     instance_double(
       'RainforestCli::Options',
-      token: nil,
+      token: 'token',
```
juliankrispel: This seems a bit hacky. If the token isn't defined, it should just be nil no?

juliankrispel: or undefined...

juliankrispel: oh, whoops, this is a spec, ignore me

juliankrispel:
@epaulet I don't see a regression test to prove that you're getting an exception when the token is not set.  It'd be great to have this for all commands. Other than that lgtm :+1:

epaulet:
You shouldn't get an exception actually. It'll just validate what it can without server data in that case.

juliankrispel:
@epaulet shouldn't it give you a message that you need to specify a token, I had this with all the other commands

README.md
```diff
@@ -62,6 +68,14 @@ You may also specify a custom test title or file name.
 rainforest new "My Awesome Title"
 ```
 
+Validate your tests for syntax and correct RFML ids for embedded tests.
+Use the `--token` options or `RAINFOREST_API_TOKEN` environment variable
+to validate your tests against server data as well.
+
+```bash
+rainforest validate
```
juliankrispel: awesome thanks

PR #142, opened 2016-06-17
Refactor/go resources

DyerSituation:
Retrieving folders, browsers, and sites and printing them in the cli.
To do: format results in readable table, write specs


resources.go
```diff
@@ -0,0 +1,69 @@
...
+	"gopkg.in/urfave/cli.v2"
+)
+
+var get getResponse
+
+type resourceParams struct {
+	Tags []string `json:"tags"`
+}
+
+func createResource(c *cli.Context, resourceType string) {
```
epaulet: This function name is confusing. Should be something like `fetchResource`

resources.go
```diff
@@ -0,0 +1,69 @@
...
+	params := makeBody(c)
```
epaulet: Don't need params in these requests

resources.go
```diff
@@ -0,0 +1,69 @@
...
+	var table [][]string
+	switch resourceType {
+	case "Folders":
+		table = getFolders(params)
+	case "Sites":
+		table = getSites(params)
+	default: //"Browsers"
```
epaulet: Let "Browsers" be an actual case and let the default either panic or return an error.

resources.go
```diff
@@ -0,0 +1,69 @@
...
+		Tags: c.StringSlice("tags"),
+	}
+}
+
+func getFolders(params *resourceParams) (tableData [][]string) {
+	var resBody *foldersResp
+	data := get.getRequest("https://app.rainforestqa.com/api/1/folders.json?page_size=100")
+	json.Unmarshal(data, &resBody)
+	tableData = resBody.TableSlice()
+	return tableData
```
epaulet: it's more efficient to return pointers in this case

epaulet: nvm this is a slice

PR #148, opened 2016-06-22
Specs for resources and resource structs

DyerSituation:
created the specs for resource fetching


getResource_test.go
```diff
@@ -5,25 +5,38 @@ import (
...
+		panic(err)
+	}
+	slice1 := []string{"firefox", "Mozilla Firefox"}
+	slice2 := []string{"ie11", "Microsoft Internet Explorer 11"}
 	expectedTable := [][]string{slice1, slice2}
-	actualTable := resBody.TableSlice()
+	matrixTestHelper(t, browsersRespBody, expectedTable)
+}
 
+func matrixTestHelper(t *testing.T, resBody returnTable, expectedTable [][]string) {
```
shosti: Can you use [`reflect.DeepEqual()`](https://golang.org/pkg/reflect/#DeepEqual) instead of manually checking?

Also, `matrixTestHelper` isn't a very descriptive name w.r.t. what it's actually checking; I like to use "narrative" names for test functions like `checkTablesAreEqual` or something like that.

getResource_test.go
```diff
@@ -0,0 +1,123 @@
...
+	for i, actualrow := range actualTable {
+		for j, actualColumn := range actualrow {
+			if expectedTable[i][j] != actualColumn {
+				t.Errorf("Unexpected matrix entry. Expected %s, got %s", expectedTable[i][j], actualColumn)
+			}
+		}
+	}
+}
+
+var fakeFolderByte = []byte(`[
```
shosti: In general, don't name variables with the type (it's redundant and distracts from the purpose). Something like `fakeFolderResp` would probably make more sense.

getResource_test.go
```diff
@@ -0,0 +1,123 @@
+package main
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestFoldersTableSlice(t *testing.T) {
+	var foldersResBody *foldersResp
```
shosti: It's a little unclear what you're actually testing here. It looks like you're mostly testing the built-in `json.Unmarshal` functionality, which is fine as a sanity check (just to double-check that your types are sane), but isn't really that valuable overall. You should really try to mostly test your actual application logic; I think a "higher-level" test would probably make more sense.

shosti:
btw, you should name files with underscores, i.e. `get_resource_test.go` and whatnot.

Resource_test.go
```diff
@@ -5,34 +5,36 @@ import (
...
+		},
 	}
-	slice1 := []string{"firefox", "Mozilla Firefox"}
-	slice2 := []string{"ie11", "Microsoft Internet Explorer 11"}
-	expectedTable := [][]string{slice1, slice2}
-	matrixTestHelper(t, browsersRespBody, expectedTable)
+	for _, tcase := range testCases {
+		err := json.Unmarshal(tcase.json, tcase.resource)
+		if err != nil {
+			panic(err)
```
shosti: This should be a `t.Fatal` instead of a `panic` (so you get nice error reporting). In general you should almost never `panic` in Go.

Resource_test.go
```diff
@@ -5,34 +5,36 @@ import (
...
-	if err != nil {
-		panic(err)
+func TestResourcesTableSlice(t *testing.T) {
+	var testCases = []struct {
+		resource      returnTable
+		json          []byte
+		expectedTable [][]string
+	}{
+		{
+			resource:      new(foldersResp),
```
shosti: I don't think you need to actually include this in the table; instead you can put a temporary var within the `for` loop and before unmarshalling.

resource_structs_test.go
```diff
@@ -0,0 +1,126 @@
...
+	if expectedLen, actualLen := len(want), len(got); expectedLen != actualLen {
+		t.Errorf("Wrong number of matrix rows. Expected %d, got %d", expectedLen, actualLen)
+	}
+	if !reflect.DeepEqual(want, got) {
+		wantB, _ := json.Marshal(want)
+		gotB, _ := json.Marshal(got)
+		t.Log("want:")
+		t.Logf("\t%+v", string(wantB))
+		t.Log("got =")
+		t.Errorf("\t%+v", string(gotB))
```
shosti: This can all be one line I think: i.e. 

```go
t.Errorf("TableSlice returned wrong result, wanted: %+v, got: %+v", want, got)
```

shosti:
I looked/though about the code some more and I have some more general comments.

# Testing

In general when writing tests, it's a good idea to think in terms of "Given X,
when I do Y, Z should happen." (X is your setup/test input, Y is the code under
test, and Z represents your assertions.) It's important that your X, Y, and Z
are on the same "level", i.e. unit-test level, integration level, etc, so that
the test is properly focused and isolated. Unit tests tend to look something
like "When I plug X into function Y, it should return/create Z"; an integration
test, on the other hand, looks at the bigger picture, in this case something
like "Given that my folder API endpoint returns X, when I run getFolders (the Y)
it should print out a table that looks like Z" (setup/assertions tend to be
quite a bit trickier for integration tests than for unit tests).

The problem with these tests is that the input is appropriate for an integration
test while your test is actually a unit test. The Y in this situation is
`GetSlice`, which "inputs" a struct and "returns" a 2-dimensional table (the
fact that it's a method instead of a function doesn't really matter). But you're
using a JSON string as an input to your test, which really belongs in an
integration test instead.

The fix is easy: use manually-created structs as your test inputs instead of
JSON strings. Besides making it easier to read, it also will make your tests
shorter and therefore less error-prone. Ideally, you should also probably have a
few integration tests as well if possible (they'll be harder to do right; we can
have a chat later if you want).

One other thing: don't worry too much about duplication when writing tests. It's
extremely important that your tests are clear and readable (your test should
make it possible for a reader to tell how the function is supposed to work
without too much difficulty), so extremely DRY code is not always the best
thing. Your test is much better now that you renamed the helper function to
`checkTablesEqual`, but I'm not actually sure it's necessary at all: I'd
probably inline the function in each of the tests for readability.

(Side note on testing with Go: when people talk about "using interfaces for
testing", they're usually talking about either the setup or the
assertions. Interfaces allow you to stub/mock the parts of your code you *don't*
want to test, as well as allowing you to check some things that would be hard
without them. I'd highly recommend
[this talk](https://www.youtube.com/watch?v=ndmB0bj7eyw) for more details on
idiomatic Go testing.)

# Code

## Naming

Naming functions/variables right is extremely important for keeping code
readable. Names should indicate a thing's purpose, rather than the details of
how it works. `TableSlice`, for instance, is a bad name for a couple of reasons:

1. It gives no indication of what the return value's purpose is. (I had to read
   through the entire rest of your code to figure out how it was used.)
2. It includes the type, which is redundant with the type signature. In general,
   functions/vars shouldn't include type information in the name unless it's for
   a very specific reason (i.e. `NewRequest` and `FormatBool` make sense because
   the type is essential to what the functions do).
   
Personally I think `makeDisplayTable` or something along those lines would be
much more descriptive: it actually tells you that the table's purpose is to be
displayed somewhere else.

## Abstraction/Interfaces

You're creating an interface across different web responses for the `TableSlice`
method. The more I think about it, the more I'm convinced that that's the wrong
level of abstraction to be working at. Responses from different endpoints are
fundamentally different creatures from a data perspective; trying to treat them
as the same thing (which you're doing by creating an interface across them)
really doesn't buy you much. (As evidence, think about how many lines of code
you're currently saving with the interface: almost none. The fact that you have
to completely re-implement almost the same function 3 times is a major
indication that something is off.)

I think there are two perfectly reasonable approaches that would improve the
code a lot:

### Option A

Abandon interfaces for this part of the code and have separate functions for
each data type (e.g. `foldersDisplayTable()`, `sitesDisplayTable()`,
etc.). That's a perfectly reasonable approach for code this simple, and really
wouldn't lead to much more duplication than you currently have.
   
### Option B
   
Abstract at a different level. If you look at the current implementations of
`TableSlice`, the only things they have in common are that each element has an
`ID` and a `Description`. So the right way to abstract, IMO, is to give each
returned element an `idDescriptioner` interface (we talked about this
earlier). You'd have to refactor your types somewhat and create separate structs
for the individual elements (which would be the things that conform to the
interface); for instance:

```go
type browsersResp struct {
	AvailableBrowsers []browser `json:"available_browsers"`
}
```

and:

```go
type sitesResp []site
```

(and move the innards to `site` and `browser` structs respectively). Then the
individual structs would have `ID()` and `Description()` methods, and you'd have
a single `displayTable(elems []idDescriptioner)` method.

Go programmers often avoid this kind of "micro-abstraction" if the duplication
isn't crazy (I'd personally probably go with Option A instead), but it might be
informative for learning more about the Go type system.

resources.go
```diff
@@ -15,38 +13,38 @@ type resourceParams struct {
 	Tags []string `json:"tags"`
 }
 
-func fetchResource(c *cli.Context, resourceType string) {
-	var table [][]string
+func fetchResource(resourceType string, web resourceGetter) (table [][]string) {
```
shosti: I actually think the original implementation is better...

http_client.go
```diff
@@ -27,8 +21,8 @@ func postRequest(url string, body []byte) []byte {
 	return data
 }
 
-func (r getResponse) getRequest(url string) []byte {
-	req, _ := http.NewRequest("GET", url, nil)
+func getRequest(url string) []byte {
+	req, _ := http.NewRequest("GET", baseURL+"/"+url, nil)
```
epaulet: we shouldn't ignore this error.

rainforest-cli.go
```diff
@@ -1,14 +1,18 @@
...
 
 	"gopkg.in/urfave/cli.v2"
 )
 
 var apiToken string
+var baseURL = "https://app.rainforestqa.com/api/1"
+var out io.Writer = os.Stdout
 
 func main() {
+
```
epaulet: Why extra newlines?

resources.go
```diff
@@ -2,20 +2,49 @@ package main
...
+	tableData = resBody.displayTable()
+	return tableData
+}
+
+func getSites() [][]string {
+	var tableData [][]string
+	var resBody *sitesResp
+	data := getRequest("sites.json")
+	fmt.Println("DATA")
+	fmt.Println(string(data))
```
epaulet: these two lines probably shouldn't be here.

http_client.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func addAuthHeaders(req *http.Request) {
+	req.Header.Add("CLIENT_TOKEN", apiToken)
+	req.Header.Add("Content-Type", "application/json")
+}
+
+func postRequest(url string, body []byte) []byte {
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(body))
+	checkAPIErr(err)
```
epaulet: This isn't really an API error.

epaulet: FWIW I would panic here because if this error exists, the code is wrong, not the client or the API.

resources.go
```diff
@@ -1,61 +1,33 @@
...
-	data := get.getRequest("https://app.rainforestqa.com/api/1/sites.json")
-	json.Unmarshal(data, &resBody)
-	tableData = resBody.TableSlice()
-	return tableData
+func printSites() {
+	var table [][]string
+	var resBody sitesResp
+	getSites("sites.json", &resBody)
+	table = resBody.displayTable()
+	printResource("Folders", table)
```
epaulet: These are sites, not folders (add a spec for this)

resources.go
```diff
@@ -1,61 +1,33 @@
...
-	data := get.getRequest("https://app.rainforestqa.com/api/1/clients.json")
-	json.Unmarshal(data, &resBody)
-	tableData = resBody.TableSlice()
-	return tableData
+func printBrowsers() {
+	var table [][]string
+	var resBody browsersResp
+	getBrowsers("clients.json", &resBody)
+	table = resBody.displayTable()
+	printResource("Folders", table)
```
epaulet: Add spec for this too

epaulet:
:ship: when green

PR #152, opened 2016-06-29
Remove yourfav cli library

DyerSituation:


resources_test.go
```diff
@@ -5,54 +5,76 @@ import (
...
-	defer ts.Close()
-	baseURL = ts.URL
+	return ts
+}
 
-	out = &bytes.Buffer{}
-	defer func() {
-		out = os.Stdout
-	}()
+func runErrorTest(resource string, t *testing.T) {
```
epaulet: `resource` is not being used in this function.

resources_test.go
```diff
@@ -99,22 +118,18 @@ func TestBrowsersFolders(t *testing.T) {
...
+func TestPrintBrowsersApiError(t *testing.T) {
+	sitesResp := `{"error": "This is a bad thing"}`
+	ts := newTestServer("/clients.json", sitesResp, 600, t)
+	defer ts.Close()
+	baseURL = ts.URL
+	out = &bytes.Buffer{}
+	defer func() {
+		out = os.Stdout
+	}()
+	runErrorTest("Browsrs", t)
```
epaulet: This shouldn't be passing with this typo. Make sure this fails once you fix the `runErrorTest` method to use the string you pass in. (Currently it's not used)

resources_test.go
```diff
@@ -5,54 +5,76 @@ import (
...
+		if r.URL.Path != path {
+			t.Errorf("fetchRource hit wrong endpoint (wanted %v but got %v)", path, r.URL.Path)
 		}
-		w.Write([]byte(siteResp))
+		w.WriteHeader(statusCode)
+		w.Write([]byte(resp))
 	}))
-	defer ts.Close()
-	baseURL = ts.URL
+	return ts
```
epaulet: I wonder if we can just return without assigning to a variable

resources.go
```diff
@@ -16,7 +16,7 @@ func printSites() {
 	getSites("sites.json", &resBody)
 	table = resBody.displayTable()
 	printResource("Site", table)
-}
+}
```
epaulet: Empty trailing space 🙀

PR #153, opened 2016-06-29
Pull API token from environment

epaulet:


rainforest-cli.go
```diff
@@ -22,11 +22,22 @@ func parseCommands() []string {
...
+
 func main() {
 	commands := parseCommands()
 	command := commands[0]
-	flag.StringVar(&apiToken, "token", "no token", "THIS IS THE USAGE")
+
+	flag.StringVar(&apiToken, "token", "", "API token. You can find your account token at https://app.rainforestqa.com/settings/integrations")
 	flag.Parse()
+
+	if len(apiToken) == 0 {
```
DyerSituation: just wondering, why do you check length and not just compare apiToken to and empty string? like `apiToken == ""`

epaulet: Both ways are idiomatic. The compile probably compiles it to the same thing either way anyway for efficiency.

Usually I use length for existence vs an empty string to actually say "I want this to be an empty string".

http://stackoverflow.com/questions/18594330/the-best-way-to-test-for-an-empty-string-in-go

PR #155, opened 2016-06-30
Add option to wait for existing run

### LINKED ISSUE #98: Feature request: attach to existing run?
Right now (afaik) there are 2 options for running tests: in the background (the default) or in the foreground (`--fg`). It would be nice if it were possible to "attach" to an existing run (i.e. like `--fg` but for a run that's already in progress--it would exit immediately with the corresponding exit code for runs that were already done).  That way, we could set off a run at the beginning of our CI process and "attach" after the unit tests finish, so the unit tests and the Rainforest tests run in parallel.

@epaulet thoughts?

shosti:
Closes #98

@epaulet this would make it easier to run our unit tests in parallel
with RF tests. What do you think?

### DIFF
--- a/.rvmrc
+++ b/.rvmrc
@@ -1,1 +1,0 @@
-rvm use 2.3.0@rainforest-cli --create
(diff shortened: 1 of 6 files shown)


shosti:
No idea what's up with the build, specs are passing (@epaulet can you look when you get a chance?)

epaulet:
Probably just needs to be rebuilt, but we should document this change in the README as well.

shosti:
@epaulet OK everything fixed I think. Can I get a review?

epaulet:
LGTM. Although we should probably figure out an easier way to use this feature other than reading standard out.

shosti:
@epaulet any ideas? Can't really think of anything that doesn't break compatibility...

IMO the "right" unixy way is to print just the run ID to stdout (that way it can be piped/stored sanely). But since all our logging goes to stdout that really isn't an option...

epaulet:
Yeah. For now, I think this is fine.

PR #156, opened 2016-07-01
[WIP] Refactor/go runs

DyerSituation:


runner.go
```diff
@@ -21,8 +22,9 @@ func createRun() {
 }
 
 func makeParams() *runParams {
+	slicedTags := strings.Split(tags, ",")
```
epaulet: Will this work if you have tags separated by a space and comma? `kind, of, like, this`

runner.go
```diff
@@ -3,31 +3,30 @@ package main
...
+func createRun() {
+	params := makeParams()
 	resBody := postRun(params)
 
 	fmt.Println(resBody)
 }
 
-func makeParams(c flagParser) *runParams {
+func makeParams() *runParams {
+	slicedTags := strings.Split(tags, ",")
```
epaulet: Will this work if you have tags separated by a space and comma? kind, of, like, this

I think the API takes care of it so not sure if we need to worry or not... probably makes sense to take care of it in both places just in case.

DyerSituation: no, it doesn't work that way as it is now. something to work on, or not a priority?

rainforest-cli.go
```diff
@@ -8,25 +8,33 @@ import (
 
 var apiToken string
```
shosti: These variable lists look a lot cleaner if you do multiple assignment, e.g.:
```go
var (
  foo int
  bar = "hello"
)
```
etc.

rainforest-cli.go
```diff
@@ -8,25 +8,33 @@ import (
...
+var smartFolderID int
+var siteID int
+var tags string
+
 var baseURL = "https://app.rainforestqa.com/api/1"
 var out io.Writer = os.Stdout
 
-func parseCommands() []string {
+func parseCommands(arguments []string) []string {
 	commands := make([]string, 0, 5)
```
shosti: why capacity 5? seems weird...

rainforest-cli.go
```diff
@@ -8,25 +8,33 @@ import (
...
-	for i := len(os.Args) - 1; i > 0; i-- {
-		if os.Args[i][0] != '-' {
-			commands = append(commands, os.Args[i])
-			os.Args = append(os.Args[:i], os.Args[i+1:]...)
+	for i := len(arguments) - 1; i > 0; i-- {
```
shosti: Why is this reverse iterating like this?

rainforest-cli.go
```diff
@@ -8,25 +8,33 @@ import (
...
 	return commands
 }
 
 func main() {
-	commands := parseCommands()
+	commands := parseCommands(os.Args)
 	command := commands[0]
 
-	flag.StringVar(&apiToken, "token", "", "API token. You can find your account token at https://app.rainforestqa.com/settings/integrations")
+	flag.StringVar(&apiToken, "token", apiToken, "API token. You can find your account token at https://app.rainforestqa.com/settings/integrations")
```
shosti: In our other Go projects we have flag vars as global variable, which I think looks cleaner. I.e. instead of this, you'd have a top level variable: 

```go
var apiToken= flag.String("token", "", ...)
```

rainforest-cli.go
```diff
@@ -8,25 +8,33 @@ import (
...
 
+var smartFolderID int
+var siteID int
+var tags string
+
 var baseURL = "https://app.rainforestqa.com/api/1"
 var out io.Writer = os.Stdout
 
-func parseCommands() []string {
+func parseCommands(arguments []string) []string {
```
shosti: Personally I think it would be a lot cleaner to parse all arguments in a helper function, e.g. `parseArgs`, that doesn't actually mutate `os.Args`. It would also make testing easier. So separate the args into two slices (commands and flags) and parse the flags using `flag.CommandLine.Parse(flags)` instead of `flag.Parse()`.

runner_test.go
```diff
@@ -1,48 +1,57 @@
...
-func (f fakeFlagParser) StringSlice(s string) []string {
-	switch s {
-	case "tags":
-		return []string{"foo", "bar"}
-	default:
-		panic(fmt.Sprintf("fakeFlagParser does not expect argument: %s", s))
-	}
+func newTestPostServer(expectedBody string, resp string, statusCode int, t *testing.T) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, _ := ioutil.ReadAll(r.Body)
```
shosti: Check the `err` here and panic if it's not `nil`.

runner_test.go
```diff
@@ -1,48 +1,57 @@
...
+		flagParams := string(body)
+		if flagParams != expectedBody {
```
shosti: Checking the actual body string seems possibly fragile, and prevents you from checking individual fields. Instead, I would probably pass in a function to `newTestPostServer`, like `check func([]byte)`, so that the actual condition handling is handled in the outer function.

runner_test.go
```diff
@@ -1,48 +1,57 @@
...
-	}
+func TestRunBySmartFolder(t *testing.T) {
+	expectedBody := `{"tags":[""],"smart_folder_id":700}`
+	sitesResp := "Post Request Successful"
+	ts := newTestPostServer(expectedBody, sitesResp, 200, t)
+	defer ts.Close()
+	baseURL = ts.URL
+	tempOsArgs := os.Args
+	os.Args = []string{"rainforest-cli", "-smart_folder_id=700", "run"}
+	main()
```
shosti: Testing `main` like this is a bit weird and not a great idea in general (except for testing super-high-level stuff like exit handling). Instead, you should move as much logic as possible out of `main` and test the sub-functions.

rainforest-cli.go
```diff
@@ -2,32 +2,50 @@ package main
...
+)
 
-func parseCommands() []string {
-	commands := make([]string, 0, 5)
-	for i := len(os.Args) - 1; i > 0; i-- {
-		if os.Args[i][0] != '-' {
-			commands = append(commands, os.Args[i])
-			os.Args = append(os.Args[:i], os.Args[i+1:]...)
+/*
+	parseCommands reverse iterates through slice of arguments to pick out all of the commands.
```
epaulet: s/parseCommands/parseArgs

runner.go
```diff
@@ -3,31 +3,34 @@ package main
...
 
 	fmt.Println(resBody)
 }
 
-func makeParams(c flagParser) *runParams {
+func makeParams() *runParams {
+	if testIDs != "" {
+		return &runParams{Tests: testIDs}
+	}
+	slicedTags := strings.Split(tags, ",")
```
epaulet: can you strip leading and trailing whitespace from tags here?

PR #158, opened 2016-07-08
Add junit output for rainforest runs

briancolby:
Must be used in foreground mode.  Uses the run id and client token to
query the rainforest api after a rainforest run has completed to build
up pass/fail information about the run.  For failed test cases uses the
"notes" given by approved testers who voted no to provide failure
reasons.


briancolby:
Didn't quite mean to open this just yet.

PR #159, opened 2016-07-12
[WIP] Refactor: Go runs

DyerSituation:
Tests can be run in foreground. Example output:

```
Run 78949 is queued and is 0% complete
Run 78949 is validating and is 0% complete
Run 78949 is in_progress and is 0% complete
...
Run 78949 is in_progress and is 81% complete
Run 78949 is in_progress and is 81% complete
Run 78949 is in_progress and is 87% complete
...
Run 78949 is in_progress and is 100% complete
Run 78949 is now complete and has failed
The detailed results are available at https://app.rainforestqa.com/runs/78949
```

options also added.

TODO:
Specs

### DIFF
--- a/http_client.go
+++ b/http_client.go
@@ -10,6 +10,8 @@ import (
 
 var client = new(http.Client)
 
+type resourceResponse interface{}
+
 func checkAPIErr(err error) {
 	if err != nil {
 		log.Fatalf("API error: %v. Please contact Rainforest support.", err)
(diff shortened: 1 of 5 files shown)


rainforest-cli.go
```diff
@@ -7,13 +7,21 @@ import (
...
+	siteID              int
+	tags                string
+	testIDs             string
+	baseURL                       = "https://app.rainforestqa.com/api/1"
+	out                 io.Writer = os.Stdout
+	runTestInBackground bool
+
+	crowd               string
+	conflict            string
+	browsers            string //[]string
```
epaulet: What does this comment mean?

runner.go
```diff
@@ -65,5 +88,40 @@ func postRun(params *runParams) (resBody *runResponse) {
...
+
+		getRun(strconv.Itoa(runID), &response)
+
+		isFinalState := response.StateDetails.IsFinalState
+		state := response.State
+		currentPercent := response.CurrentProgress.Percent
+
+		if !isFinalState {
+			fmt.Printf("Run %v is %v and is %v%% complete\n", runID, state, currentPercent)
+			//TODO implement failfast
```
epaulet: Don't forget to do this.

shosti:
Build is failing, plz check.

http_client.go
```diff
@@ -87,3 +87,16 @@ func getBrowsers(url string, resBody *browsersResp) []byte {
 	checkAPIErr(err)
 	return data
 }
+
+func getRun(runID string, resBody *runResponse) {
+	req, err := http.NewRequest("GET", baseURL+"/runs/"+runID, nil)
+	checkPanicError(err)
```
shosti: This shouldn't panic, as it's bound to happen sometimes.

shosti: oh wait, nvm

rainforest-cli.go
```diff
@@ -7,13 +7,22 @@ import (
...
+	out                 io.Writer = os.Stdout
+	runTestInBackground bool
+	failfast            bool
+
+	crowd               string
+	conflict            string
+	browsers            string
+	description         string
+	environmentID       int
+	testingNoForeGround bool
```
shosti: What is this?

http_client.go
```diff
@@ -87,3 +87,16 @@ func getBrowsers(url string, resBody *browsersResp) []byte {
...
+	addAuthHeaders(req)
+	res, err := client.Do(req)
+	checkAPIErr(err)
+	checkResponse(res)
+	data, err := ioutil.ReadAll(res.Body)
+	checkAPIErr(err)
+	err = json.Unmarshal(data, &resBody)
+	checkAPIErr(err)
```
shosti: Also it seems like most of this logic (request + error checking + unmarshaling) is common to all kinds of requests and should be abstracted away.

http_client.go
```diff
@@ -10,6 +10,8 @@ import (
 
 var client = new(http.Client)
 
+type resourceResponse interface{}
```
shosti: what is this? Shouldn't use `interface{}` without a good reason--what's the reasoning here?

shosti: OIC. I guess that makes sense, although I wouldn't typecast it like here (just use `interface{}` directly in the method that does the work).

runner_test.go
```diff
@@ -188,4 +200,58 @@ func TestRunByTestID(t *testing.T) {
 		createRun()
 		ts.Close()
 	}
+	runTestInBackground = tempTestInBackground
+}
+func TestRunInForeground(t *testing.T) {
+	var percent int
+	tempTestInBackground := runTestInBackground
+	runTestInBackground = false
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
shosti: This testing method looks good overall, but there are a couple minor issues:

1. It's very long (to the point where it's hard to read all at once). Try moving some stuff into helper functions.
2. I haven't run it, but it seems like the delay might happen in the test. It's never a good idea to have `sleep`s in your tests since it slows them down unnecessarily; you might want to move the polling interval into a global variable or something.
3. You're only testing the "happy path", and even then you have very few assertions (for quite a lot of work). There are at least 3 cases that are important:

 a. "Happy case" (test runs and passes), which you test pretty well.
 b. Test fails and fail-fast == false (which means that it keeps going even after it fails and then exits with 0)
 c. Test fails and fail-fast == true (which means it quits immediately with 0 exit code)

I'm not sure you have to do (b), but (c) is pretty important I'd say.

ukd1:
@epaulet @DyerSituation this is on pause, right?

epaulet:
Yeah. We can reopen when we pick it up.

PR #162, opened 2016-07-14
Junit output

briancolby:
Add junit output for rainforest runs
-----------------------------------

Must be used in forground mode.  Uses the run id and client token to
query the rainforest api after a rainforest run has completed to build
up pass/fail information about the run.  For failed test cases uses the
"notes" given by approved testers who voted no to provide failure
reasons.

Generating a useful Junit output from the API takes at least two GETs.  A hit on `run` and a hit on `tests`, the former to get information about the total number of tests, the amount of time it was run etc, while the latter is needed to get slightly more detail for each test case efficiently. The tests endpoint allows the parser to determine the amount of time the test took to complete (typical content of Junit reports).

Where the parser becomes slightly heavier is with failed test cases, where it needs to hit the API to get the specific details of each test failure.  It then uses this information to gather the valid notes on the test failure to act as a "traceback" for the test failure.  Noting the browser and the failure reason.

Here is an example of what this parsed output looks like in a CI tool like Bamboo:
image

image

This is largely accomplished by the second class I added `JunitOutputter`.  This class does the heavy lifting of organizing the JSON data returned from the rainforest API into an XML report, using `XmlBuilder` to handle all of the XML.

I added unit tests for all the classes I modified, and added new specs for the classes I added, and tried to follow the coding style of the modules as I was editing them.

A couple of concerns, The return from the `tests` endpoint is paginated. My approach of specifying that the page returned contain as many test cases as the run will only function if there either is no upper limit on the number of tests which can be returned, or if the run being reported on contains fewer test cases than that limit.  This logic may need to be made more complicated.

Another is that this Junit output generates results tracking 1 to 1 with the number of rainforest test cases.  When issuing a rainforest run you can ask that a test case be run against many different browsers and environments. And I do not feel like this information is being properly reflected in the Junit output.

I think what I have here is already a pretty long PR (luckily most of the lines are test fixtures though) and it provides at least the minimum of getting the number of passing and failing test cases from a rainforest run along with the given reason for test failures. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ Popular command line options are:
 - `--embed-tests` - Use with `rainforest export` to export your tests without extracting the
 steps of an embedded test.
 - `--test-folder /path/to/directory` - Use with `rainforest [new, upload, export]`. If this option is not provided, rainforest-cli will, in the case of 'new' create a directory, or in the case of 'upload' and 'export' use the directory, at the default path `./spec/rainforest/`.
-- `--junit` - Create a junit xml report file with the specified name.  Must be run in foreground mode, or with the report command. Uses the rainforest
+- `--junit-file` - Create a junit xml report file with the specified name.  Must be run in foreground mode, or with the report command. Uses the rainforest
 api to construct a junit report.  This is useful to track tests in CI such as Jenkins or Bamboo.
 - `--run-id` - Only used with the report command.  Specify a past rainforest run by ID number to generate a report for.
 
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ rainforest browsers
 
 To generate a junit xml report for a test run which has already completed
 ```bash
-rainforest report --run_id <run_id> --junit rainforest.xml
+rainforest report --run-id <run-id> --junit rainforest.xml
 ```
 
 ## Options
@@ -194,7 +194,7 @@ steps of an embedded test.
 - `--test-folder /path/to/directory` - Use with `rainforest [new, upload, export]`. If this option is not provided, rainforest-cli will, in the case of 'new' create a directory, or in the case of 'upload' and 'export' use the directory, at the default path `./spec/rainforest/`.
 - `--junit` - Create a junit xml report file with the specified name.  Must be run in foreground mode, or with the report command. Uses the rainforest
 api to construct a junit report.  This is useful to track tests in CI such as Jenkins or Bamboo.
-- `--run_id` - Only used with the report command.  Specify a past rainforest run by ID number to generate a report for.
+- `--run-id` - Only used with the report command.  Specify a past rainforest run by ID number to generate a report for.
 
 
 #### Specifying Test IDs
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ Run all tests with tag 'run-me' and abort previous in-progress runs.
 rainforest run --tag run-me --fg --conflict abort
 ```
 
+Run all in the foreground and generate a junit xml report.
+
+```bash
+rainforest run all --fg --junit rainforest.xml
+```
+
 ###### Creating and Managing Tests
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -181,6 +192,9 @@ Popular command line options are:
 - `--embed-tests` - Use with `rainforest export` to export your tests without extracting the
 steps of an embedded test.
 - `--test-folder /path/to/directory` - Use with `rainforest [new, upload, export]`. If this option is not provided, rainforest-cli will, in the case of 'new' create a directory, or in the case of 'upload' and 'export' use the directory, at the default path `./spec/rainforest/`.
+- `--junit` - Create a junit xml report file with the specified name.  Must be run in foreground mode, or with the report command. Uses the rainforest
+api to construct a junit report.  This is useful to track tests in CI such as Jenkins or Bamboo.
+- `--run_id` - Only used with the report command.  Specify a past rainforest run by ID number to generate a report for.
```
epaulet: I think `--run-id` would be more consistent with the rest of the multi-word flags.

lib/rainforest/cli/junit_outputter.rb
```diff
@@ -0,0 +1,73 @@
...
+    attr_reader :builder, :client
+
+    def initialize(token, run, tests)
+      @client = HttpClient.new token: token
+      @json_run = run # JSON containing the results of /1/runs/{run_id}.json
+      @json_tests = tests # JSON containing the results of /1/runs/{run_id}/tests.json
+      @builder = Builder::XmlMarkup.new( :indent => 2)
+    end # end initialize
+
+    def dump_summary
```
epaulet: Is this used anywhere?

epaulet:
Thanks for the contribution @briancolby

I added a couple of comments inline. But I also noticed that this fails when I enter a path that doesn't exist as the value for `--junit`. ie:

```bash
$ rainforest report --junit ./foo/bar/baz.xml --run_id 80091
I, [2016-07-15T14:32:19.880609 #30542]  INFO -- : Generating JUNIT report for 80091 : ./foo/bar/test.xml
/Users/Edward/Desktop/RainforestQA/rainforest-cli/lib/rainforest/cli/reporter.rb:50:in `initialize': No such file or directory @ rb_sysopen - ./foo/bar/test.xml (Errno::ENOENT)
	from /Users/Edward/Desktop/RainforestQA/rainforest-cli/lib/rainforest/cli/reporter.rb:50:in `open'
	from /Users/Edward/Desktop/RainforestQA/rainforest-cli/lib/rainforest/cli/reporter.rb:50:in `report'
	from /Users/Edward/Desktop/RainforestQA/rainforest-cli/lib/rainforest/cli.rb:42:in `start'
	from bin/rainforest:7:in `<main>'
```

README.md
```diff
@@ -181,6 +192,9 @@ Popular command line options are:
 - `--embed-tests` - Use with `rainforest export` to export your tests without extracting the
 steps of an embedded test.
 - `--test-folder /path/to/directory` - Use with `rainforest [new, upload, export]`. If this option is not provided, rainforest-cli will, in the case of 'new' create a directory, or in the case of 'upload' and 'export' use the directory, at the default path `./spec/rainforest/`.
+- `--junit` - Create a junit xml report file with the specified name.  Must be run in foreground mode, or with the report command. Uses the rainforest
```
epaulet: We have another tag used to specify files called `--import-variable-csv-file`, so perhaps `--junit-file` may be more consistent and clear.

dmehra:
Thanks @epaulet, @briancolby is away for a few days, feel free to make the changes on the branch if you'd like, or he will pick next week.

briancolby:
I brought the command line option names into style with the others following your recommendations, and added logic to create directories which are specified but do not exist.

`dump_summary` was a leftover function for testing before I factored out the actual writing of output to the Reporter class (allowing me to more easily leverage StringIO for testing).

PR #183, opened 2016-08-20
Don't try to parameterize values that are already set

epaulet:
Currently we don't check if the parameters for `file.download` and `file.screenshot` are already in place. We try to replace them all and give an error.


ukd1:
@epaulet lgtm

PR #191, opened 2016-09-19
Delete existing generators with same name when uploading new ones

epaulet:


lib/rainforest_cli/csv_importer.rb
```diff
@@ -27,9 +27,19 @@ def import
       end
       raise 'Invalid schema in CSV. You must include headers in first row.' if !columns
 
-      print 'Creating custom step variable'
+      puts 'Checking for existing tabular variables.'
+      generators = http_client.get('/generators')
+      generator = generators.find { |g| g['name'] == @generator_name }
+
+      if generator
```
shosti: What happens currently?

epaulet:
I thought about that as well - but if you are using this feature, you are looking to overwrite your existing variable anyway. worst case scenario - generators are soft-deleted so they can ask us to reenable.

I think I would rather change behavior on the client end than behavior in the API.

lib/rainforest_cli/csv_importer.rb
epaulet: If you want your variable to be named `my_var` and that already exists, it creates `my_var_1`.

shosti: Can you add a new CLI option (`--overwrite` or something)? Don't want anyone to trigger this by accident.

epaulet: @shosti just added an overwrite option

lib/rainforest_cli/options.rb
```diff
@@ -42,14 +42,18 @@ def initialize(args)
...
+        opts.on('--import-variable-csv-file FILE', 'Import tabular step variables; CSV data') do |value|
           @import_file_name = value
         end
 
-        opts.on('--import-variable-name NAME', 'Import step variables; Name of variable (note, will be replaced if exists)') do |value|
+        opts.on('--import-variable-name NAME', 'Import tabular step variables; Name of variable') do |value|
           @import_name = value
         end
 
+        opts.on('--overwrite-variable', 'Import tabular step variables: overwrite existing variable if desired variable name is taken') do
```
shosti: should this be overwrite-variables?

epaulet: We are only updating one variable at a time so I think the singular is fine.

PR #199, opened 2016-10-14
Add tolerance for server errors

epaulet:
This retries not just on Timeouts, but also on non-200 status codes

### DIFF
--- a/lib/rainforest_cli/http_client.rb
+++ b/lib/rainforest_cli/http_client.rb
@@ -38,7 +38,7 @@ module RainforestCli
             return JSON.parse(response.body)
           elsif options[:retries_on_failures] && response.code >= 500
             delay = retry_delay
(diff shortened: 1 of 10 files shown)


lib/rainforest_cli/http_client.rb
```diff
@@ -13,81 +13,58 @@ def initialize(options)
...
-            Kernel.sleep 5
+          response = Http::Exceptions.wrap_exception do
+            HTTParty.send(method, url, { body: body, headers: headers, verify: false })
+          end
+
+          if response.code.between?(200, 299)
+            return JSON.parse(response.body)
+          elsif options[:retries_on_failures] && response.code >= 500
+            delay = retry_delay
+            logger.warn("HTTP request was unsuccessful. URL: #{url}. Status: #{response.code}")
```
curtis-rainforestqa: any reason for the params here and not below?

PR #204, opened 2016-10-27
Don't upload test meta data to description anymore

epaulet:
We no longer need to save test meta-data in the notes section, just regular notes.

### DIFF
--- a/lib/rainforest_cli/test_parser.rb
+++ b/lib/rainforest_cli/test_parser.rb
@@ -26,8 +26,7 @@ module RainforestCli::TestParser
       @test.browsers = []
     end
 
-    TEST_DATA_FIELDS = [:start_uri, :title, :site_id].freeze
-    STEP_DATA_FIELDS = [:redirect].freeze
+    TEST_DATA_FIELDS = [:start_uri, :title, :site_id, :tags, :browsers].freeze
     CSV_FIELDS = [:tags, :browsers].freeze
 
     def process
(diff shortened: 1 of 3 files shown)


lib/rainforest_cli/test_parser.rb
```diff
@@ -26,8 +26,7 @@ def initialize(file_name)
       @test.browsers = []
     end
 
-    TEST_DATA_FIELDS = [:start_uri, :title, :site_id, :browsers].freeze
-    STEP_DATA_FIELDS = [:redirect].freeze
+    TEST_DATA_FIELDS = [:start_uri, :title, :site_id, :tags, :browsers].freeze
```
curtis-rainforestqa: is the addition of `:tags` intentional here?

lib/rainforest_cli/test_parser.rb
```diff
@@ -37,32 +36,31 @@ def process
...
 
-            # if it's supposed to be a CSV field, split and trim it
-            if CSV_FIELDS.include?(field)
-              @test[field] = @test[field].split(',').map(&:strip).map(&:downcase)
+              # if it's supposed to be a CSV field, split and trim it
+              if CSV_FIELDS.include?(matched_field)
+                @test[matched_field] = @test[matched_field].split(',').map(&:strip).map(&:downcase)
+              end
             end
           end
```
curtis-rainforestqa: any tests we should update for this?

lib/rainforest_cli/test_parser.rb
epaulet: Yeah. The only time we use `TEST_DATA_FIELDS` alone is when we concat it with `CSV_FIELDS` anyway.

lib/rainforest_cli/test_parser.rb
epaulet: Added some specs for this. I thought they already existed.

lib/rainforest_cli/test_parser.rb
curtis-rainforestqa: right on

lib/rainforest_cli/test_parser.rb
curtis-rainforestqa: NICE

PR #209, opened 2016-11-17
[Review] Groundwork for a refactor

macocha:
@epaulet 

spec: https://paper.dropbox.com/doc/Product-Narrative-Rainforest-cli-go-version-nsU4nrv5LciPqfIzNkuVk


rainforest-cli.go
```diff
@@ -1,83 +1,304 @@
...
+			Description: "Download your remote tests from Rainforest to RFML. " +
+				"You need to specify list of test IDs to download or use keyword 'all'. " +
+				"Alternatively you can use one of the filtering options.",
+			Flags: []cli.Flag{
+				cli.StringSliceFlag{
+					Name:  "tag",
+					Usage: "filter tests by `TAG`. Can be used multiple times for filtering by multiple tags.",
+				},
+				cli.StringFlag{
+					Name:  "site, site-id",
```
epaulet: I saw a few commands that now have multiple aliases. We should keep the README up to date as we change things so we don't lose track maybe?

rainforest/rainforest.go
```diff
@@ -0,0 +1,120 @@
...
+	"encoding/json"
+	"errors"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+)
+
+const (
+	currentBaseURL  = "https://app.rainforestqa.com/api/1/"
```
epaulet: In the Ruby version, this is also configurable with a ENV var - are we keeping that?

rainforest-cli.go
```diff
@@ -1,83 +1,304 @@
...
+			Aliases: []string{"r"},
+			Usage:   "Run your tests on Rainforest",
+			Action:  startRun,
+			Description: "Runs your tests on Rainforest platform. " +
+				"You need to specify list of test IDs to run or use keyword 'all'. " +
+				"Alternatively you can use one of the filtering options.",
+			ArgsUsage: "[test IDs]",
+			Flags: []cli.Flag{
+				cli.StringSliceFlag{
+					Name:  "tag",
```
macocha: Easiest way would be to move it to constant and use in both places.

rainforest-cli.go
```diff
@@ -1,83 +1,304 @@
...
-	flag.StringVar(&crowd, "crowd", "", "Crowd to run test with. Enter `default` or `on_premise_crowd`")
-	flag.StringVar(&conflict, "conflict", "", "Handling of runs in progress. (A) Abort: only abort runs in the same environment as your new run. (B) Abort All.")
-	flag.StringVar(&browsers, "browsers", "", "Browsers to test against. enter in a comma separated list")
-	flag.StringVar(&description, "description", "", "An arbitrary string to associate with the run")
-	flag.IntVar(&environmentID, "environment_id", 0, "Use a specific environment for this run")
-	flag.BoolVar(&runTestInBackground, "bg", false, "run test in background")
-	flag.BoolVar(&runTestInBackground, "fail-fast", true, "run test in background")
-	flag.CommandLine.Parse(flags)
+	app := cli.NewApp()
+	app.Usage = "CLI client for Rainforest QA services - http://rainforestqa.com"
```
ukd1: plz fix this -- ``Rainforest QA CLI - https://www.rainforestqa.com/``

rainforest-cli.go
```diff
@@ -1,83 +1,304 @@
...
-		}
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:   "token, t",
+			Usage:  "API token. You can find it at https://app.rainforestqa.com/settings/integrations",
+			EnvVar: "RAINFOREST_API_TOKEN",
+		},
+		cli.IntFlag{
+			Name:   "threads",
+			Usage:  "Used to customize the amount of threads to use when making HTTP requests.",
```
ukd1: threads?

rainforest-cli.go
ukd1: or kill the aliases in the new version

rainforest-cli.go
```diff
@@ -1,83 +1,304 @@
...
+						"Can be used multiple times to run against multiple browsers.",
+				},
+				cli.StringFlag{
+					Name:  "environment-id",
+					Usage: "run your tests using specified `ENVIRONMENT`. Otherwise it will use your default one.",
+				},
+				cli.StringFlag{
+					Name:  "crowd",
+					Value: "default",
+					Usage: "run your tests using specified `CROWD`. Available choices are: default or on_premise_crowd.",
```
ukd1: can you add a note to talk to CSM about this; not everyone has this

rainforest-cli.go
```diff
@@ -1,83 +1,304 @@
...
+				},
+				cli.StringFlag{
+					Name: "conflict",
+					Usage: "use the abort option to abort any runs in the same environment or " +
+						"use the abort-all option to abort all runs in progress.",
+				},
+				cli.BoolTFlag{
+					Name: "fg, foreground",
+					Usage: "results in the foreground - rainforest-cli will not return until the run is complete. " +
```
ukd1: ``show results in ...``

rainforest/rainforest.go
```diff
@@ -0,0 +1,120 @@
...
+	if res.StatusCode >= 200 && res.StatusCode < 300 {
+		return nil
+	}
+	// Otherwise we return error
+	// TODO: We might add some better error handling here, like parsing error response.
+	return errors.New("RF API Error")
+}
+
+// Do sends out the request to the API and unpacks JSON response to the out variable.
+func (c *Client) Do(req *http.Request, out interface{}) (*http.Response, error) {
```
ukd1: can you also add a version header? both the _version_ and the build?

this how to do build (it's called version here....);
https://github.com/rainforestapp/tadameat/blob/develop/main.go#L14 - make it a global
https://github.com/rainforestapp/tadameat/blob/develop/circle.yml#L9 - set when build
https://github.com/rainforestapp/tadameat/blob/develop/main.go#L31 - display it out (though, you should add to a header)

rainforest/resources.go
```diff
@@ -0,0 +1,115 @@
...
+// GetDescription returns the Folder name
+func (f Folder) GetDescription() string {
+	return f.Title
+}
+
+// GetFolders returns a slice of Folders (their names and IDs) which are available
+// for filtering RF tests.
+func (c *Client) GetFolders() ([]Folder, error) {
+	// Prepare request
+	req, err := c.NewRequest("GET", "folders.json?page_size=100", nil)
```
ukd1: does this work with > 100 folders?

rainforest/resources.go
```diff
@@ -0,0 +1,115 @@
...
+
+// GetDescription returns the Site name
+func (s Site) GetDescription() string {
+	return s.Name
+}
+
+// GetSites fetches sites available to use during the RF runs.
+func (c *Client) GetSites() ([]Site, error) {
+	// Prepare request
+	req, err := c.NewRequest("GET", "sites.json", nil)
```
ukd1: why do these all have ``.json``?

rainforest/runner.go
```diff
@@ -0,0 +1,77 @@
...
+		return runStatus, err
+	}
+	_, err = c.Do(req, &runStatus)
+	if err != nil {
+		return runStatus, err
+	}
+
+	// Here we check if we successfully created a run in RF, if not then bail with an error.
+	if runStatus.ID == 0 {
+		return runStatus, errors.New("Failed to create a run.")
```
ukd1: prolly show the error. 

also, correct way to check is status code, not this way

runner.go
```diff
@@ -1,128 +1,188 @@
...
-		return &runParams{Tests: testIDsSlice}
-	}
-	tags = strings.TrimSpace(tags)
-	var slicedTags []string
-	if tags != "" {
-		slicedTags = strings.Split(tags, ",")
-		for i, slice := range slicedTags {
-			slicedTags[i] = strings.TrimSpace(slice)
+		if !isFinalState {
+			fmt.Printf("Run %v is %v and is %v%% complete\n", runID, state, currentPercent)
```
ukd1: can you use ``log`` here not ``fmt``? it'll output a nice parsable log with time

ukd1:
can y'all link to the spec plz?

rainforest/resources.go
epaulet: No. Maybe we should try to get the next 100 as well if the first request returned 100 results.

rainforest/resources.go
epaulet: I don't think it does anything right now. I think we've thought to do `.json` on everything before so we can get json 500 errors instead of html 500s, but that doesn't work either.

rainforest/rainforest.go
macocha: Done!

epaulet:
Gonna merge this to the main go branch so that @macocha and I can split the work without a chain of branches. Things still to do (will add to other PR as well):

- [ ] Kill or document flag aliases
- [ ] Fetch more than 100 resources

PR #212, opened 2016-11-23
Go runner specs

epaulet:


resources.go
```diff
@@ -36,8 +36,15 @@ func printResourceTable(resourceName string, data [][]string) {
...
+// resourceAPI is part of the API connected to available resources
+type resourceAPI interface {
+	GetFolders() ([]rainforest.Folder, error)
+	GetBrowsers() ([]rainforest.Browser, error)
+	GetSites() ([]rainforest.Site, error)
+}
+
 // printFolders fetches and prints out the available folders from the API
-func printFolders(c *cli.Context) error {
+func printFolders(c *cli.Context, api resourceAPI) error {
```
epaulet: I think it makes more sense to pass the pointer into the functions.

epaulet:
@macocha I branched off of this already so feel free to merge in if everything is good, but we probably shouldn't push anymore to this branch.

PR #214, opened 2016-11-24
CSV upload

macocha:
Still need to add:
- [x] tests
- [x] pre-run csv upload

@epaulet

### DIFF
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -13,14 +13,6 @@ import (
 const (
 	// Version of the app in SemVer
 	version = "2.0.0"
-
-	// Run status poll interval
-	runStatusPollInterval = time.Second * 5
-
-	// Batch size (number of rows) for tabular var upload
-	tabularBatchSize = 50
-	// Concurent connections when uploading CSV rows
-	tabularConcurency = 1
 )
 
 var (
@@ -31,8 +23,16 @@ var (
 	// Rainforest API client
 	api *rainforest.Client
 
-	// default output for printing
-	out io.Writer = os.Stdout
+	// default output for printing resource tables
+	tablesOut io.Writer = os.Stdout
+
+	// Run status polling interval
+	runStatusPollInterval = time.Second * 5
+
+	// Batch size (number of rows) for tabular var upload
+	tabularBatchSize = 50
+	// Concurent connections when uploading CSV rows
+	tabularConcurency = 1
 )
 
 // notImplemented is a placholder function for actions that are not yet implemented.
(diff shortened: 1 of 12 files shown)


rainforest-cli.go
```diff
@@ -26,8 +23,16 @@ var (
...
+	// default output for printing resource tables
+	tablesOut io.Writer = os.Stdout
+
+	// Run status polling interval
+	runStatusPollInterval = time.Second * 5
+
+	// Batch size (number of rows) for tabular var upload
+	tabularBatchSize = 50
+	// Concurent connections when uploading CSV rows
+	tabularConcurency = 1
```
epaulet: Typo, but also is there any way to change this / do we want to allow that?

rainforest-cli.go
```diff
@@ -126,12 +131,12 @@ func main() {
 					Usage: "use the abort option to abort any runs in the same environment or " +
 						"use the abort-all option to abort all runs in progress.",
 				},
-				cli.BoolTFlag{
+				cli.BoolFlag{
 					Name: "fg, foreground",
```
epaulet: Leaving this as false by default for now?

rainforest/tabularvars.go
```diff
@@ -0,0 +1,174 @@
...
+	"strconv"
+	"time"
+)
+
+// Generator is a type representing generators which can be used as variables in RF tests.
+// They can be builtin or uploaded by customer as a tabular variable.
+type Generator struct {
+	ID           int               `json:"id,omitempty"`
+	Name         string            `json:"name,omitempty"`
+	CreationDate time.Time         `json:"created_at, omitempty"`
```
epaulet: This field isn't used at all, is it?

tabularvars.go
```diff
@@ -0,0 +1,175 @@
...
+	return b
+}
+
+// rowUploadWorker is a helper worker which reads batch of rows to upload from rows chan
+// and pushes potential errors through errorsChan
+func rowUploadWorker(api tabularVariablesAPI, generator rainforest.Generator,
+	columns []string, rowsChan <-chan [][]string, errorsChan chan<- error) {
+	for rows := range rowsChan {
+		error := api.AddGeneratorRowsFromTable(generator, columns, rows)
+		errorsChan <- error
```
epaulet: The other rows will still upload. Should we stop the uploads immediately on all of the workers?

rainforest/tabularvars.go
macocha: We are not using it in our CLI, but it's included in the library, as someone might want to get to that. 😃

rainforest-cli.go
macocha: Oh yeah, but I think we should switch it around and have `bg, background` flag.

rainforest-cli.go
macocha: It's more for us to have it in a easy accessible place if we decide to change it later on.

tabularvars.go
macocha: As decided on slack, we're leaving it as is, because they'll get killed anyways when main routine exits.

PR #219, opened 2016-12-01
[Ready] git trigger in go

macocha:
@epaulet :)

### DIFF
--- a/gittrigger.go
+++ b/gittrigger.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"bytes"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+const gitTriggerString = "@rainforest"
+
+type gitTrigger struct {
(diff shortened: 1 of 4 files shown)


gittrigger.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil
+}
+
+func (g gitTrigger) checkTrigger() bool {
+	shouldTrigger := strings.Contains(g.LastCommit, g.Trigger)
+	return shouldTrigger
+}
+
+func (g gitTrigger) getTags() []string {
+	tagRegex, _ := regexp.Compile("#([\\w_-]+)")
```
epaulet: is there an extra `\` here?

macocha: Yeah, it's to escape `\` so it's `\w` in regex, I guess I could avoid this escaping using backtick, but oh, well. ;)

PR #226, opened 2016-12-05
[Review] RFML Support in rainforest-lib + validate cmd & rm cmd

macocha:


macocha:
Tests for this will be done after we have complimentary writer, as it will make test case creation easier.

rainforest-cli.go
```diff
@@ -205,8 +205,9 @@ func main() {
 		{
 			Name:      "validate",
 			Usage:     "Validate your RFML tests",
-			ArgsUsage: " ",
-			Description: "Validate your tests for syntax and correct RFML ids for embedded tests. " +
+			ArgsUsage: "[path to RFML file]",
```
epaulet: Should add this to the README.

rainforest-cli.go
```diff
@@ -236,10 +237,10 @@ func main() {
 		},
 		{
 			Name:        "rm",
-			Usage:       "Remove an RFML test locally and remotely",
+			Usage:       "Remove an RFML test locally and remotely (if token is specified)",
```
epaulet: I think we should force them to have a token. Otherwise people are encouraged to delete their tests locally but not remotely. If they just want to remove a local file, they don't need us to do that for them :D

macocha: Good point, I don't know why it somehow slipped my mind, that you can actually remove file yourself... lol...

rainforest-cli.go
macocha: Added!

PR #228, opened 2016-12-06
[Review] Go file uploads

epaulet:
This based off of the branch in #227


epaulet:
There's a bunch of changes taken from #226 that aren't really part of this PR but shows up in the diffs.

PR #232, opened 2016-12-09
[Review] Add self-updating to the client

macocha:
This adds `update` command which can be used to manually trigger updates and switch between channels and auto-updates that are ran async while cli runs.  
I've also set up 3 release channels:
 - `stable` for normal stable releases, that gets pushed from tagged releases in form of `v2.1.3`
 - `beta` for beta releases, gets pushed from tag on any branch in form of `v2.1.3-beta.1`
- `dev` bleeding edge channel, gets pushed after every merge to develop - used to hopefully run RF tests against it.  

All of the needed keys are set as env vars in Circle.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -26,3 +26,36 @@ dependencies:
 test:
   override:
     - cd $BUILD && go test -v -race $(glide novendor)
+
+deployment:
+  stable:
+    branch: master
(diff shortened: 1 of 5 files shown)


macocha:
cc @ukd1 @epaulet

PR #236, opened 2016-12-15
[Review] Download command

epaulet:


rainforest/tests.go
```diff
@@ -197,6 +197,23 @@ func (c *Client) GetRFMLIDs() (TestIDMappings, error) {
 	return testResp, nil
 }
 
+// GetTest gets a test from RF specified by the given test ID
+func (c *Client) GetTest(testID int) (*RFTest, error) {
+	req, err := c.NewRequest("GET", "tests/"+strconv.Itoa(testID), nil)
```
ziahamza: its probably better to just use fmt.Sprintf(...) in these cases @epaulet , but just a minor nitpick

rfml.go
```diff
@@ -476,6 +464,120 @@ func uploadRFMLFilesInDirectory(rfmlDirectory string) error {
 	return nil
 }
 
+func downloadRFML(c cliContext) error {
+	testDirectory := c.String("test-folder")
```
ziahamza: any specific reason to use `test-folder` here @epaulet  ... was this part of the new spec?

rfml.go
```diff
@@ -476,6 +464,120 @@ func uploadRFMLFilesInDirectory(rfmlDirectory string) error {
...
+}
+
+/*
+	Helper Functions
+*/
+
+func prepareTestDirectory(testDir string) (string, error) {
+	absTestDirectory, err := filepath.Abs(testDir)
+	if err != nil {
+		return "", cli.NewExitError(err.Error(), 1)
```
ziahamza: nuke `NewExitError` from here, as we already do it from upstream: https://github.com/rainforestapp/rainforest-cli/pull/236/files#diff-0dae6029cce0003a807265a154d65843R471

rfml.go
```diff
@@ -476,6 +464,120 @@ func uploadRFMLFilesInDirectory(rfmlDirectory string) error {
...
+	for i := 0; i < rfmlDownloadConcurrency; i++ {
+		go downloadRFTestWorker(testIDChan, errorsChan, testChan)
+	}
+
+	for i := 0; i < len(testIDs); i++ {
+		select {
+		case err = <-errorsChan:
+			return err
+		case test := <-testChan:
+			test.UnmapBrowsers()
```
ziahamza: Wouldnt it be better to have a single function to do all the necessary conversion, like `test.Init(mappings)`  rather then separate calls to `UnmapBrowsers`/`UmarshalElements`, as more need to be added later on @macocha @epaulet

rfml.go
```diff
@@ -476,6 +464,120 @@ func uploadRFMLFilesInDirectory(rfmlDirectory string) error {
...
+	mappings, err = api.GetRFMLIDs()
+	if err != nil {
+		return cli.NewExitError(err.Error(), 1)
+	}
+
+	var testIDs []int
+	if len(c.Args()) > 0 {
+		var testID int
+		for _, arg := range c.Args() {
+			testID, err = strconv.Atoi(arg)
```
ziahamza: Probably better to use `strconv.ParseUint` here: https://golang.org/pkg/strconv/#example_ParseUint

As we cant have negative testIds

rfml.go
```diff
@@ -476,6 +464,120 @@ func uploadRFMLFilesInDirectory(rfmlDirectory string) error {
 	return nil
 }
 
+func downloadRFML(c cliContext) error {
```
ziahamza: specs for this?

rfml.go
epaulet: This is in the old CLI too actually. It defaults to `spec/rainforest`.

rfml.go
epaulet: Actually, that makes sense. Similar to the current `PrepareToUploadFromRFML`.

rainforest/tests.go
epaulet: I only did it this way because I thought it was more efficient. I'll take a look at our string concatenation in a diff PR since it's all over the project.

epaulet:
I addressed the comments - gonna merge so we can start live testing tomorrow.

PR #237, opened 2016-12-16
[Go] Retrieve more than 100 folders with folders command

epaulet:


rainforest/resources.go
```diff
@@ -36,10 +40,40 @@ func (c *Client) GetFolders() ([]Folder, error) {
...
+	if totalPagesHeader == "" {
+		return nil, fmt.Errorf("Missing X-Total-Pages header in HTTP Response")
+	}
+
+	totalPages, err := strconv.Atoi(totalPagesHeader)
+	if err != nil {
+		return nil, err
+	}
+
+	if totalPages > 1 {
```
ziahamza: we can nuke the condition, the `for` statement should be enough

rainforest/resources_test.go
```diff
@@ -4,29 +4,69 @@ import (
...
 	"net/http"
 	"reflect"
+	"regexp"
+	"strconv"
 	"testing"
 )
 
 func TestGetFolders(t *testing.T) {
 	setup()
-	defer cleanup()
```
ziahamza: why move cleanup from the end ... was `defer` not working properly @epaulet ?

PR #239, opened 2016-12-22
[Review] Add custom-url behavior

epaulet:
Just recently realized that this was missing


epaulet:
We should, but we currently don't do that in the old CLI so I didn't bother with it yet.

PR #240, opened 2016-12-23
[Review] Go Automatic file uploads

epaulet:


rfml.go
```diff
@@ -326,34 +326,69 @@ func uploadSingleRFMLFile(filePath string) error {
...
-		return err
-	}
-	if parsedTest.TestID == 0 {
-		// Create new test
+
+	testID, ok := mappings.MapRFMLIDtoID()[parsedTest.RFMLID]
+	if ok {
+		parsedTest.TestID = testID
+	} else {
+		// Create an empty test
```
macocha: Do we need to go through empty test route if we are uploading just a single file?

rfml.go
```diff
@@ -446,7 +482,25 @@ func uploadRFMLFilesInDirectory(rfmlDirectory string) error {
 	// And here we update all of the tests
 	testsToUpdate := make(chan *rainforest.RFTest, len(parsedTests))
 	for _, testToUpdate := range parsedTests {
-		testToUpdate.PrepareToUploadFromRFML(mappings)
+		testID, ok := mappings.MapRFMLIDtoID()[testToUpdate.RFMLID]
+		if ok {
+			testToUpdate.TestID = testID
```
macocha: this is actually done in `PrepareToUploadFromRFML`

rainforest/rfml.go
```diff
@@ -263,3 +269,120 @@ func (r *RFMLWriter) WriteRFMLTest(test *RFTest) error {
...
+					return "", err
+				}
+				filePath = filepath.Join(usr.HomeDir, filePath[2:])
+			} else if test.RFMLPath == "" {
+				return "", fmt.Errorf("Cannot parse relative file path %v for RFML test %v. RFMLPath field cannot be blank.", filePath, test.RFMLID)
+			} else {
+				rfmlDirectory := filepath.Dir(test.RFMLPath)
+				filePath = filepath.Join(rfmlDirectory, filePath)
+			}
+
```
macocha: if you are using `filepath.Abs` you don't really need to do `~/` and `rfmlDirectory` joins above. Just set `rfmlDirectory` as a working directory and Abs should do all of the magicks for you.

rfml.go
epaulet: The issue I had was that I needed a test ID to upload embedded files.

rfml.go
epaulet: Only if it exists. I had to change the logic around so that I always had the test ID before uploading embedded files.

PR #241, opened 2016-12-31
[Review] Download filters

epaulet:


rainforest-cli.go
```diff
@@ -290,12 +290,12 @@ func main() {
 					Name:  "tag",
 					Usage: "filter tests by `TAG`. Can be used multiple times for filtering by multiple tags.",
 				},
-				cli.StringFlag{
+				cli.IntFlag{
```
epaulet: Figured it made sense to use integers flags here so that if they try to put a string as the folder, the CLI will tell them that it's incorrect usage:

```
Edward$ go build && ./rainforest-cli download --folder-id lol
Incorrect Usage.

NAME:
   rainforest-cli download - Download your remote Rainforest tests to RFML

USAGE:
   rainforest-cli download [command options] [test IDs]

DESCRIPTION:
   Download your remote tests from Rainforest to RFML. You need to specify list of test IDs to download or use keyword 'all'. Alternatively you can use one of the filtering options.

OPTIONS:
   --tag TAG                                  filter tests by TAG. Can be used multiple times for filtering by multiple tags.
   --site SITE-ID, --site-id SITE-ID          filter tests by a specific site. You can see a list of your SITE-IDs with sites command. (default: 0)
   --folder FOLDER-ID, --folder-id FOLDER-ID  filter tests by a specific folder. You can see a list of your FOLDER-IDs with folders command. (default: 0)
   --test-folder PATH                         PATH at which to save all the downloaded tests. (default: "./spec/rainforest/") [$RAINFOREST_TEST_FOLDER]
   --embed-tests                              download your tests without extracting the steps of an embedded test.
```

PR #246, opened 2017-01-30
add flags shuffle

macocha:
This is a hack around the `urfave/cli` "issue" with global flags supplied as command parameters, not being recognized.


epaulet:
Looks good, but I don't think we should skip making a few unit tests for this as it could break everything if it didn't work properly.

PR #259, opened 2017-04-11
[WIP] Fetch all tests

epaulet:
There's currently an issue where the CLI only fetches the first handful of tests returned by the API. This should fix that problem.


epaulet:
This should be merged into develop. Closing.

PR #267, opened 2017-05-31
Empty browsers and tags attributes fix

epaulet:
If you leave
`# browsers:`

blank, you will get the following error:
"You have selected at least one unauthorized browers."

Because the CLI will send the following attribute to the server
`{ "browsers": [""] }`

This PR should fix that (and a similar issue for tags), but not omitting the `browsers` or `tag` attributes in the body of the request, and instead sends an empty array if the browser/tag list is empty or omitted.

### DIFF
--- a/rainforest/tests_test.go
+++ b/rainforest/tests_test.go
@@ -2,12 +2,12 @@ package rainforest
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"reflect"
 	"strconv"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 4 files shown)


rainforest/tests_test.go
```diff
@@ -2,10 +2,12 @@ package rainforest
 
 import (
 	"encoding/json"
+	"io/ioutil"
```
curtis-rainforestqa: what's this used for?

epaulet: Used here: https://github.com/rainforestapp/rainforest-cli/pull/267/files#diff-18a89673e7ce64a1cb0c89479bbda183R232

curtis-rainforestqa: ah

PR #275, opened 2017-06-09
tests and functionality to init runs with run groups and view run groups

sondhayni-rfqa:


rainforest-cli.go
```diff
@@ -132,6 +132,10 @@ func main() {
 					Usage: "filter tests by `TAG`. Can be used multiple times for filtering by multiple tags.",
 				},
 				cli.StringFlag{
+					Name:  "run-group-id, run-group",
```
epaulet: Let's just use `run-group-id` because I think the other aliases are being used mostly for backwards compatibility reasons.

PR #279, opened 2017-06-21
Only use the first 16 characters of a test's title in the rfml path

epaulet:


rfml.go
```diff
@@ -663,7 +663,13 @@ func sanitizeTestTitle(title string) string {
 
 	// replace all non-alphanumeric characters with an underscore
 	rep := regexp.MustCompile(`[^[[:alnum:]]`)
-	return rep.ReplaceAllLiteralString(title, "_")
+	title = rep.ReplaceAllLiteralString(title, "_")
+
+	if len(title) > 16 {
+		return title[:16]
+	}
```
magni-: Technically, this doesn't need to be in an if-test, right?

epaulet: It does because if the length is less than 16, you access parts of the slice that are out of bounds. I made the same assumption when writing this myself 😅

magni-: Go 🎉

PR #287, opened 2017-07-29
RF-2642 Run local tests only

shosti:
This is the "dumb" way, but shouldn't require any changes on the
Rainforest end. It should be ready to go but I haven't tested
yet. @epaulet mind giving it a look?

### DIFF
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -140,12 +140,12 @@ func main() {
 					Usage: "filter tests by `TAG`. Can be used multiple times for filtering by multiple tags.",
 				},
 				cli.StringSliceFlag{
-					Name:  "dont-execute",
+					Name:  "exclude",
 					Usage: "Don't execute test specified by `FILE`. Can be used multiple times for specifying muliple files.",
 				},
 				cli.StringSliceFlag{
-					Name:  "execute",
-					Usage: "Execute test specified by `FILE` even if execute: false is specified.",
+					Name:  "force-execute",
+					Usage: "Execute test specified by `FILE` even if execute: false is specified. Can be used multiple times for specifying multiple files.",
 				},
 				cli.StringFlag{
 					Name:  "run-group-id",
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -131,6 +131,10 @@ func main() {
 				"Alternatively you can use one of the filtering options.",
 			ArgsUsage: "[test IDs]",
 			Flags: []cli.Flag{
+				cli.BoolTFlag{
+					Name:  "f, files",
+					Usage: "only run local tests. Specify any number of files or folders to run as the last arguments.",
+				},
 				cli.StringSliceFlag{
 					Name:  "tag",
 					Usage: "filter tests by `TAG`. Can be used multiple times for filtering by multiple tags.",
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -248,7 +248,9 @@ func main() {
 					EnvVar: "RAINFOREST_TEST_FOLDER",
 				},
 			},
-			Action: validateRFML,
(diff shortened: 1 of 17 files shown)


rainforest/rainforest.go
```diff
@@ -57,7 +57,12 @@ func NewClient(token string, debug bool) *Client {
 		baseURL, _ = url.Parse(currentBaseURL)
 	}
 
-	return &Client{client: http.DefaultClient, BaseURL: baseURL, ClientToken: token, LastResponseHeaders: http.Header{}, DebugFlag: debug}
+	return &Client{client: http.DefaultClient, BaseURL: baseURL, clientToken: token, LastResponseHeaders: http.Header{}, DebugFlag: debug}
+}
+
+// ClientToken returns the API authentication token for the client.
+func (c *Client) ClientToken() string {
```
epaulet: Why not export the field? To protect it from being changed?

rfml.go
```diff
@@ -22,27 +22,122 @@ type fileParseError struct {
...
+		if seenPaths[filePath] {
+			continue
+		}
+		seenPaths[filePath] = true
+		test, err := readRFMLFile(filePath)
+		if err != nil {
+			return nil, err
+		}
+		if tags == nil || anyMember(test.Tags, tags) {
+			parsedTests = append(parsedTests, parsedTest{filePath, test})
```
epaulet: why don't we just attacked filePath to the RFMLPath field on the test?

rainforest/rainforest.go
shosti: So that it could be overwritten in tests. But it breaks backward-compatibility I realized, maybe not the best idea after all... (Or we should just change `Client` to be an interface, which would make testing in other modules quite a bit easier.)

rfml.go
shosti: Yeah that does seem better :smile: . Should we just nuke `parsedTest`?

rainforest/rainforest.go
epaulet: Ah. Setting the field value in the test manually has issues with race conditions?

shosti: No, you just can't do it with a "fake" client (which I wanted to use).

rfml.go
epaulet: Yeah, I think it would be cleaner if we didn't have duplicate fields on different levels of structs.

shosti:
@epaulet I went for an extremely large refactor to help with testing. Please let me know what you think--I'm not super attached to it and could go either way.

shosti:
I tested locally and it *mostly* does the right thing. 

I realized there's a problem, though: it requires you to specify the embedded tests, and then it runs them! Oops! We'll have to figure out a solution (maybe `--embedded-tests` or something?).

shosti:
I decided not to go with the big refactor after all 😄 .

Should be gtg I think...

rainforest-cli.go
```diff
@@ -131,10 +131,22 @@ func main() {
...
+				cli.BoolTFlag{
+					Name:  "f, files",
+					Usage: "run local tests specified by `FILES or FOLDERS`",
+				},
 				cli.StringSliceFlag{
 					Name:  "tag",
 					Usage: "filter tests by `TAG`. Can be used multiple times for filtering by multiple tags.",
 				},
+				cli.StringSliceFlag{
+					Name:  "dont-execute",
```
ukd1: also noticed none of these changes are in the README.md. Plz fix.

shosti:
@ukd1 better?

shosti:
Created a run and it worked! https://app.rainforestqa.com/runs/179461

Created using `./rainforest-cli run --browser chrome_1440_900 -f ~/src/rainforest/regenwald/spec/rainforest --tag smoke-suite`

PR #289, opened 2017-08-25
RF-2789 Return false exit status for unknown flags

jbarber:
So that giving bad flags doesn't appear the same as having a passing
test suite

https://rainforest.atlassian.net/browse/RF-2789

No idea how to test this?

```
$ ./rainforest-cli --foo
2017/08/26 00:19:29 Unknown argument
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,3 +288,30 @@ Email [help@rainforestqa.com](mailto:help@rainforestqa.com) if you're having tro
 4. Commit your changes (`git commit -am 'Add some feature'`)
 5. Push to the branch (`git push origin my-new-feature`)
 6. Create new Pull Request
+
+## Release process
+
+Check the `circle.yml` for the latest, but currently merging to master will build and deploy to the following Equinox channels:
+
+Tag                             | Channels
+--------------------------------|-------------
+No tag                          | dev
+vX.Y.Z-alpha.N or vX.Y.Z-beta.N | beta, dev
+vX.Y.Z                          | stable, beta, dev
+
+Development + release process is:
+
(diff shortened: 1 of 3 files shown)


epaulet:
While we're at it, why not also add the `CommandNotFound` function? Or is there already a reasonable default?

jbarber:
@epaulet 
```
$ ~/usr/bin/rainforest asdasd
2017/08/29 12:34:04 No help topic for 'asdasd'
$ echo $?
3
```

So I think it's handled already?

jbarber:
@epaulet is there anything else that needs to be done to release this apart from merging?

epaulet:
After we merge, we will need to update the version constant and the change log. Once all that's done in and in the master branch, we just need to tag the final commit with the version number. That will trigger the circle build that releases the new version. Then we just need to update the Github change logs.

README.md
```diff
@@ -288,3 +288,29 @@ Email [help@rainforestqa.com](mailto:help@rainforestqa.com) if you're having tro
 4. Commit your changes (`git commit -am 'Add some feature'`)
 5. Push to the branch (`git push origin my-new-feature`)
 6. Create new Pull Request
+
+## Release process
+
+Check the `circle.yml` for the lastest, but currently merging to master will build and deploy to the following Equinox channels:
```
epaulet: Typo on `latest`

jbarber:
@epaulet can you re-review the changes I've made here. The CHANGELOG reformating is to make it more consistent and make it easier to automatically update the GH releases page (see #292)

PR #292, opened 2017-09-03
Automate adding GH release notes when a stable release is made

jbarber:
Not sure if the `GITHUB_AUTH_TOKEN` required for access to the requests API has the appropriate perms, I guess we'll see when we try to release...


epaulet:
@jbarber did this end up not working?

jbarber:
Not needed

PR #293, opened 2017-09-04
RF-2852 Backoff if the API returns an error

jbarber:
When waiting for a run to complete, instead of failing if we get an API
error, backoff and retry upto 5 times.

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -222,6 +222,12 @@ func monitorRunStatus(c cliContext, runID int) error {
 			return nil
 		}
 
+		// If we've had too many errors, give up
+		if backoff >= 5 {
+			msg := fmt.Sprintf("Can not get run status after %d attempts, giving up", backoff)
+			return cli.NewExitError(msg, 1)
+		}
+
 		// If we hit an error, wait longer before retrying
 		if err != nil {
 			backoff++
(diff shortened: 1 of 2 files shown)


runner.go
```diff
@@ -202,61 +202,78 @@ func filterExecuteTests(tests []*rainforest.RFTest, tags []string, forceExecute,
...
-		// Wait for tick
-		<-t.C
-		newStatus, err := api.CheckRunStatus(runID)
-		if err != nil {
-			resChan <- statusWithError{status: newStatus, err: err}
-			return
-		}
+// Pow provides exponentiation for ints.
+// Only in Go would I find myself having to write this
+func Pow(x, y int) int {
```
epaulet: I don't think we need this to be an exported function in the main package.

runner.go
```diff
@@ -202,61 +202,78 @@ func filterExecuteTests(tests []*rainforest.RFTest, tags []string, forceExecute,
...
+			if status.Result != "passed" {
+				return cli.NewExitError("", 1)
+			}
+
+			return nil
+		}
+
+		// If we hit an error, wait longer before retrying
+		if err != nil {
+			backoff++
```
epaulet: We should print something out saying that we are going to pause briefly so that it's clear why we are waiting for potentially 2 minutes between requests.

runner.go
```diff
@@ -202,61 +202,78 @@ func filterExecuteTests(tests []*rainforest.RFTest, tags []string, forceExecute,
...
+			// Reset backoff
+			backoff = 1
+		}
+		// If we've had too many errors, give up
+		if backoff > 5 {
+			return cli.NewExitError(err.Error(), 1)
+		}
 
-	return nil
+		time.Sleep(runStatusPollInterval * time.Duration(Pow(backoff, 2)))
```
epaulet: This appears to be a maximum of 125 seconds. I wonder if 5 is too high or whether we'll run into issues with the process timing out in some builds...

PR #294, opened 2017-09-08
RF-3270 CLI Parity for RF2

shosti:


rainforest/resources.go
```diff
@@ -12,50 +12,67 @@ type Folder struct {
...
-	// Prepare request
-	req, err := c.NewRequest("GET", "folders?page_size=100", nil)
+// getPaginatedResource gets all of a resource from endpoint (using multiple
+// requests, if necessary) and returns the result.
+//
+// Because of the type system, usage is hairy. coll should be a pointer to an
+// empty slice of the appropriate type, and collect is called every time
+// resources are added to the collection. The caller should handle collecting
+// the collection.
+func (c *Client) getPaginatedResource(endpoint string, coll interface{}, collect func(interface{})) error {
```
epaulet: Instead of using a callback, can't we have `getPaginatedResource` return a blank interface and just handle the return value in the caller? Not really sure if that's any less gross though...

shosti: I tried soooo many things, couldn't get anything to work other than this :sob:. The problem is `interface{}` and `[]interface{}` work differently and it's extremely hard to move things in between.

epaulet:
Any applicable changes for the README?

PR #298, opened 2017-09-22
RF 2.0 features & disabled tests

epaulet:


rainforest/rfml.go
```diff
@@ -253,6 +278,13 @@ func (r *RFMLWriter) WriteRFMLTest(test *RFTest) error {
 		}
 	}
 
+	if test.State == "disabled" {
+		_, err = writer.WriteString("# disabled: true\n")
```
shosti: Why not `state: disabled`? (Defaulting to `enabled`?)

rainforest/rfml_test.go
```diff
@@ -85,6 +91,29 @@ func TestReadAll(t *testing.T) {
 		t.Errorf("Incorrect values for RFTest.\nGot %#v\nWant %#v", rfTest, validTestValues)
 	}
 
+	// Test is disabled
+	testText = fmt.Sprintf(`#! %v
+# title: %v
+# start_uri: %v
+# disabled: %v
```
shosti: Hard to tell, but is there a test to check that it's enabled by default?

rainforest/tests.go
```diff
@@ -45,8 +45,10 @@ type RFTest struct {
 	RFMLID      string                   `json:"rfml_id"`
 	Source      string                   `json:"source"`
 	Title       string                   `json:"title,omitempty"`
+	State       string                   `json:"state,omitempty"`
```
shosti: To double-check: this is how it works in the API, right?

rainforest/rfml.go
epaulet: For some reason I thought we decided to go with the `# disabled: boolean` route, but I can change it to go by the state.

shosti: lol there's no tech spec so I don't think we decided anything :smile:. I think going with the API way is better to avoid confusion, and maybe we should leave it out by default when downloading tests to RFML.

rainforest/tests.go
epaulet: @shosti Actually I think we need to change the API here. Sending a PUT request with the a state that the test already has returns a 400 (so if you upload a test twice without changing the state, 400). 👎

epaulet: This has been fixed 🎉

rainforest/rfml.go
```diff
@@ -124,6 +126,26 @@ func (r *RFMLReader) ReadAll() (*RFTest, error) {
 						return parsedRFTest, &parseError{lineNumStr, "Redirect value must be a valid boolean"}
 					}
 					currStepRedirect = redirect
+				case "feature_id":
+					if value == "" {
+						// If the value is empty, delete the feature
```
shosti: To clarify: this won't get triggered if `feature_id` is blank?

rainforest/tests.go
```diff
@@ -39,18 +40,38 @@ func (s TestIDMappings) MapRFMLIDtoID() map[string]int {
 	return resultMap
 }
 
+// FeatureIDInt is a wrapper the int type used in the FeatureID field of RFTest
+// that implements the json.Marshaler interface.
+type FeatureIDInt int
+
+const deletedFeatureID = -1
```
shosti: The naming is a bit strange, why not just `deleteFeature` or something?

rainforest/tests_test.go
```diff
@@ -403,6 +416,8 @@ func TestUpdateTest(t *testing.T) {
 			t.Errorf("Unexpected browsers received. Expected: [], Got: %v", bodyStr)
 		} else if !strings.Contains(bodyStr, "\"tags\":[]") {
 			t.Errorf("Unexpected tags received. Expected: [], Got: %v", bodyStr)
+		} else if !strings.Contains(bodyStr, "\"folder_id\":null") {
```
shosti: Is there already a test that it *doesn't* include this if `folder_id` is 0?

rainforest/rfml.go
epaulet: We won't reach this case if the file does not say "# feature_id:" in it.

rainforest/tests_test.go
epaulet: Yep, line 338 has that test.

PR #304, opened 2017-10-06
Add file path to validation errors

epaulet:
We were forgetting to wrap the errors properly in a couple of instances.

The error output has the file path in it for validations:
```
$ ./testing --skip-update validate
2017/10/06 13:44:38 spec/rainforest/0000097596_testing_the_cli.rfml: RFML parsing error in line 11: Each step must contain a question, with a `?`
```

### DIFF
--- a/rfml_test.go
+++ b/rfml_test.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	"github.com/rainforestapp/rainforest-cli/rainforest"
+	"github.com/urfave/cli"
 )
 
 func TestNewRFMLTest(t *testing.T) {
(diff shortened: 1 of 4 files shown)


rfml_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	"github.com/rainforestapp/rainforest-cli/rainforest"
+	"github.com/urfave/cli"
```
curtis-rainforestqa: is this used for anything?

jbarber: Go won't let you include things that are not used, it raises an error at compile time!
e.g. https://play.golang.org/p/ndFKKsMopT

PR #305, opened 2017-10-13
Add default JUnit report name

epaulet:
Also did some minor refactoring since we weren't using the one of the methods attached to the reporter struct.

### DIFF
--- a/reporter.go
+++ b/reporter.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/xml"
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


reporter.go
```diff
@@ -239,24 +216,40 @@ func createJunitTestReportSchema(runID int, tests []rainforest.RunTestDetails, a
...
+		Tests:     runDetails.TotalTests,
+		TestCases: testCases,
+		Time:      runDuration,
+	}
+
+	if runDesc := runDetails.Description; runDesc != "" {
+		report.Name = runDesc
+	} else {
+		report.Name = fmt.Sprintf("Run #%v", runDetails.ID)
+	}
```
epaulet: ^ This part is the main change. The rest is refactoring.

PR #306, opened 2017-10-16
RF-3843 GA run -f

shosti:


README.md
```diff
@@ -127,6 +127,24 @@ Download specific tests based on their id on the Rainforest dashboard
...
 
+#### Running Local RFML Tests Only
+
+If you want to run a local set of RFML files (for instance in a CI environment), use the `run -f` option:
+
+```
+rainforest run -f [FILES OR FOLDERS]
+```
+
+`run -f` accepts any number of files and folders as arguments (folders will be recursively checked for `*.rfml` files). All embedded tests must be included within `FILES OR FOLDERS`.
```
epaulet: Do you think it's clear enough that we accept pattern matched paths?

README.md
```diff
@@ -127,6 +127,24 @@ Download specific tests based on their id on the Rainforest dashboard
...
+
+There is a specific metadata option in RFML files for `run -f`: `# execute: true|false`, which indicates whether a test should be run by default (defaults to `true`). Embedded tests that are not usually run directly should specify `# execute: false`.
+
+Most options for `rainforest run` work the same way with `run -f`, with the exception of test filtering options (such as `--folder`). In addition, there are a few options that are specific to `run -f` or behave differently:
```
epaulet: This seems ambiguous. Are there any other options that do not work with `-f` or is it only the filtering options that work differently/don't work at all?

README.md
shosti: We don't accept pattern matched paths (the shell does expansion). I think "power users" will assume that.

shosti:
@epaulet better?

PR #315, opened 2017-11-02
Update to 2.8.6

shosti:


epaulet:
Should we bother documenting the change of which parameter we use for starting runs? (`folder_id` -> `feature_id`)

shosti:
The API change isn't user-facing so I don't think we need a Changelog for it.

PR #316, opened 2017-11-02
Update CLI package and handle usage errors

### LINKED ISSUE #312: Command failure does not return error code
`
  rainforest run --token $RAINFOREST_TOKEN --fg --fail-fast --description "CircleCI 
  $CIRCLE_BUILD_NUM $CIRCLE_BUILD_URL" --tag other

 Incorrect Usage.
`
echo $?
0

Should be an error.

epaulet:
Also removed old vendor directory stuff. This directory is already listed in .gitignore ever since we started using glide.

Apparently adding the usage error handler at the top level (ie `app.OnUsageError`) does not affect usage errors for specific commands.

fixes #312


rainforest-cli.go
```diff
@@ -125,10 +125,11 @@ func main() {
...
-			Aliases: []string{"r"},
-			Usage:   "Run your tests on Rainforest",
-			Action:  startRun,
+			Name:         "run",
+			Aliases:      []string{"r"},
+			Usage:        "Run your tests on Rainforest",
+			OnUsageError: onCommandUsageErrorHandler("run"),
+			Action:       startRun,
 			Description: "Runs your tests on Rainforest platform. " +
 				"You need to specify list of test IDs to run or use keyword 'all'. " +
```
jbarber: Is this indentation legit? It looks odd that the string for the `Description` field doesn't start in the same column as the other values.

epaulet: I agree that it looks weird/inconsistent but it's what gofmt gives me :man_shrugging:

PR #319, opened 2017-11-03
Revert "Revert "Update CLI package and handle usage errors""

epaulet:
Reverts rainforestapp/rainforest-cli#317


epaulet:
Deprecation error is gone:
```
$ go build -o testing *.go
$ ./testing --skip-update run --token foobar --environment-id 356 --folder 1312 --conflict abort --bg --description "foo"
2017/11/03 13:08:09 Account not found
```

epaulet:
Ideally we should add some top level integration tests for each command, but we can add that as a ToDo for now I think...

epaulet:
For historical purposes, the reason this fixed it is because the signature for the action functions changed in 1.15.0: https://github.com/urfave/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature

shosti:
So much for type safety :disappointed:

PR #320, opened 2017-11-28
Add a Dockerfile

shosti:
This will make it easier for us to use CircleCI 2.0 internally, and
might also be generally useful for CI if it's in the Docker hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.9.2
+WORKDIR /go/src/github.com/rainforestapp/rainforest-cli
+RUN curl -s https://glide.sh/get | sh
+ADD glide.yaml glide.lock ./
+RUN glide install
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.9.2
+WORKDIR /go/src/github.com/rainforestapp/rainforest-cli
+RUN curl -s https://glide.sh/get | sh
+ADD glide.yaml glide.lock ./
```
macocha: any particular reason why add just those two files, glide install and then copy over the rest?

shosti: So the layer cache doesn’t get bust every time a file changes. Not really a big deal but it was annoying me 😉.

macocha: oh good point 👍

Issue #339, opened 2018-12-05
Junit output is malformed


crabbylion:
Noticing that generating JUnit output from rainforest runs, either through the report command, or with the junit option flag on a foreground run produce inaccurate xml files. The output files are limited to 100 test cases (despite indicating the correct number of test cases on the header line.

I did check to make sure that the failed test cases werent simply hidden by the truncation to 100 test cases, this list did include at least one failed test case. As you can see it is not marked as an error and it does not include the tester comments as failure reasons.

Here is an example XML showing what I mean: 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="STAGONE #621 https://ci.int.quid.com/browse/RAIN-STAGONE-JOB1-621" tests="205" errors="0" failures="5" time="3250.954">
  <testcase name="Add Companies To Saved Search Portfolio" time="2357"></testcase>
  <testcase name="Adjust Min &amp; Max Axis Range on Scatter Plot" time="1361"></testcase>
  <testcase name="Any Mention fields in Bar Chart" time="1993"></testcase>
  <testcase name="Category Labels" time="1539"></testcase>
  <testcase name="change color by, size by, and labeling in news network view" time="2025"></testcase>
  <testcase name="Change view options in barchart on companies network" time="1937"></testcase>
  <testcase name="change view options in barchart on news network" time="1537"></testcase>
  <testcase name="Change view options in barchart on patents network" time="1524"></testcase>
  <testcase name="Change view options in histogram on companies network" time="2764"></testcase>
  <testcase name="change view options in histogram on news network" time="2748"></testcase>
  <testcase name="Change view options in histogram on patents network" time="2106"></testcase>
  <testcase name="Change view options in scatter plot in companies network" time="905"></testcase>
  <testcase name="change view options in scatter plot on news network" time="2359"></testcase>
  <testcase name="change view options in timeline of news network" time="2817"></testcase>
  <testcase name="Changing color by, size by, and labeling in Companies Network View" time="429"></testcase>
  <testcase name="Cluster Selection Filter Applied and Stays Applied with any Chart Type" time="2515"></testcase>
  <testcase name="cluster wizard delete cluster" time="1828"></testcase>
  <testcase name="cluster wizard merge clusters" time="2334"></testcase>
  <testcase name="color and size by popular choices in news network view" time="1739"></testcase>
  <testcase name="Color by popular choices in Companies Network View" time="2050"></testcase>
[...]

Issue #216, opened 2016-11-28
Test "updated_at" not exposed in GET /runs/:id


epaulet:
Since we do not expose `updated_at` on tests in GET `/runs/:id`, created JUnit test reports requires an extra HTTP requests to `/runs/:id/tests`. This is likely a waste as `created_at` should not be expensive to add to `/runs`, which would save an HTTP request making it both quicker on the client side and easier on the server side.

epaulet:
Since we also need `description` exposed in the API as well (which could have performance issues), this doesn't seem like it's worth the effort.

epaulet:
I was looking at the wrong API! We can indeed bypass the need for an extra request when creating JUnit reports by just exposing `updated_at` in the API.

PR #340, opened 2019-01-12
RF-11904 fix JUnit output

epaulet:
With the change to the run tests API to include `updated_at`, we no longer have to make an extra requests to `GET /runs/:id/tests` (see deleted comment in PR for more details).

I also added two query string parameters that were required but not present. `include_feedback` exposes failure reasons, which is necessary for creating the JUnit output. `skip_mark_as_viewed` so that the failed tests don't get marked as viewed by the GET request.

addresses #216 #339


epaulet:
The old version does not work anyway because we did some page size deprecations and some breaking API changes without updating this.

rainforest/reporter.go
```diff
@@ -68,31 +69,13 @@ func (c *Client) GetRunDetails(runID int) (*RunDetails, error) {
 		return &runDetails, err
 	}
 
-	// NOTE: This extra request is only necessary because `updated_at` is not
-	// currently exposed in the `/runs/:id` endpoint. This may change in the future:
-	// https://github.com/rainforestapp/rainforest-cli/issues/216
-	var runTests []RunTestDetails
-	url = "runs/" + strconv.Itoa(runID) + "/tests?page_size=" + strconv.Itoa(runDetails.TotalTests)
```
epaulet: This endpoint is paginated with a max page size of 100. Therefore, if there were >100 run tests, you would not get a complete set of results in your JUnit file. However, we don't need to make this request to retrieve `updated_at` anymore - this information is already present in the previous request above. And that request is NOT paginated, so you will get complete results just by removing this.

PR #343, opened 2019-01-16
Fix the stable deploy

epaulet:


epaulet:
I accidentally copied the wrong thing. This one is copied over correctly from the old `circle.yml`.

PR #348, opened 2019-04-12
RF-14049 - Multiple app support for `mobile-upload`

shanempope:


rainforest/mobile_upload.go
```diff
@@ -158,27 +170,27 @@ type SiteEnvironment struct {
 	URL           string `json:"url"`
 }
 
-func (c *Client) getSiteEnvironmentID(siteID int, environmentID int) (int, error) {
+func (c *Client) getSiteEnvironmentID(siteID int, environmentID int) (int, string, error) {
```
epaulet: Since we are returning more than just the ID, I think we should rename this to reflect that. Or have it return some sort of `SiteEnvironment` struct that contains the ID and url.

PR #349, opened 2019-04-23
RF-14590 - Increase upload app slots to 100 for `mobile-upload` command `--app-slot` flag.

shanempope:


igorfrenkel:
Will not approve as I am not on the requested reviewers list, but will give my 👍Very small change affecting a changed constant.

PR #355, opened 2019-05-06
RF-14256 Set priority on RFML

fuzzylita:
/reviewme @rainforestapp/backend @epaulet 

Did my best but honestly, serious copypasta. Please let me know if I biffed anything ❤️

### DIFF
--- a/rainforest/rfml.go
+++ b/rainforest/rfml.go
@@ -140,6 +140,14 @@ func (r *RFMLReader) ReadAll() (*RFTest, error) {
 					parsedRFTest.FeatureID = FeatureIDInt(featureID)
 				case "state":
 					parsedRFTest.State = value
+				case "priority":
+					value = strings.ToUpper(value)
+					switch value {
+					case "P1", "P2", "P3", "":
+						parsedRFTest.Priority = value
+					default:
+						return parsedRFTest, &parseError{lineNumStr, "Redirect value must be one of P1, P2, P3"}
+					}
 				case "execute":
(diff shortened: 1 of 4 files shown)


rainforest/rfml.go
```diff
@@ -140,6 +140,14 @@ func (r *RFMLReader) ReadAll() (*RFTest, error) {
 					parsedRFTest.FeatureID = FeatureIDInt(featureID)
 				case "state":
 					parsedRFTest.State = value
+				case "priority":
+					value = strings.ToUpper(value)
+					switch value {
+					case "P1", "P2", "P3", "":
+						parsedRFTest.Priority = value
+					default:
+						return parsedRFTest, &parseError{lineNumStr, "Redirect value must be one of P1, P2, P3"}
```
igorfrenkel: The error says only p1, p2, p3 acceptable, but in the case statement on line 146 you also accept "". Is this intentional?

fuzzylita: I thought I should, but was working with a seasoned Go dev to learn while doing and he suggested that it was unnecessary 😅

Happy to add!

epaulet: Well the issue is that we need a way to "unset" priority, right? Seems like empty string would be the way to go. Does the API remove priority if you send an empty string?

fuzzylita: I'll add that in a separate PR.

fuzzylita:
@jbarber priority is optional, so it should be fine.

PR #362, opened 2019-11-18
RF-16284 add `automation_and_crowd` crowd option

kruszczynski:
Adds the option

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -289,7 +289,7 @@ func (r *runner) makeRunParams(c cliContext, localTests []*rainforest.RFTest) (r
 	}
 
 	var crowd string
-	if crowd = c.String("crowd"); crowd != "" && crowd != "default" && crowd != "on_premise_crowd" && crowd != "automation" {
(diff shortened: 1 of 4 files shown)


runner.go
```diff
@@ -289,7 +289,7 @@ func (r *runner) makeRunParams(c cliContext, localTests []*rainforest.RFTest) (r
 	}
 
 	var crowd string
-	if crowd = c.String("crowd"); crowd != "" && crowd != "default" && crowd != "on_premise_crowd" && crowd != "automation" {
+	if crowd = c.String("crowd"); crowd != "" && crowd != "default" && crowd != "on_premise_crowd" && crowd != "automation" && crowd != "automation_and_crowd" {
```
grodowski: I kind of hate go for making these kind of statements the obvious default 😅 https://stackoverflow.com/questions/10485743/contains-method-for-a-slice

kruszczynski: That's why I like it 🚎

PR #370, opened 2020-03-18
RF-16455 Report RFA failures in Junit

kruszczynski:
Add support for RFA failures that are reported differently

### DIFF
--- a/rainforest/reporter.go
+++ b/rainforest/reporter.go
@@ -34,14 +34,14 @@ type RunTestBrowserDetails struct {
 
 // RunTestDetails contains details about a Run Test
 type RunTestDetails struct {
-	ID             int                     `json:"id"`
-	Title          string                  `json:"title"`
-	CreatedAt      time.Time               `json:"created_at"`
-	UpdatedAt      time.Time               `json:"updated_at"`
-	Result         string                  `json:"result"`
-	Steps          []RunStepDetails        `json:"steps"`
-	HasWispResults bool                    `json:"has_wisp_results"`
-	Browsers       []RunTestBrowserDetails `json:"browsers"`
+	ID            int                     `json:"id"`
+	Title         string                  `json:"title"`
+	CreatedAt     time.Time               `json:"created_at"`
+	UpdatedAt     time.Time               `json:"updated_at"`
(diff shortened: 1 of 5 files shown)


rainforest/reporter.go
```diff
@@ -26,14 +26,22 @@ type RunStepDetails struct {
...
-	UpdatedAt time.Time        `json:"updated_at"`
-	Result    string           `json:"result"`
-	Steps     []RunStepDetails `json:"steps"`
+	ID             int                     `json:"id"`
+	Title          string                  `json:"title"`
+	CreatedAt      time.Time               `json:"created_at"`
+	UpdatedAt      time.Time               `json:"updated_at"`
+	Result         string                  `json:"result"`
+	Steps          []RunStepDetails        `json:"steps"`
+	HasWispResults bool                    `json:"has_wisp_results"`
```
maciejgryka: Nitpick - is `Wisp` here something we could change to say e.g. `HasRfaResults`, or did that ship already sail?

kruszczynski: We can change the field name, we probably can change the API as well since it's not public

kruszczynski: Done!

rainforest/reporter.go
```diff
@@ -26,14 +26,22 @@ type RunStepDetails struct {
...
-	UpdatedAt time.Time        `json:"updated_at"`
-	Result    string           `json:"result"`
-	Steps     []RunStepDetails `json:"steps"`
+	ID            int                     `json:"id"`
+	Title         string                  `json:"title"`
+	CreatedAt     time.Time               `json:"created_at"`
+	UpdatedAt     time.Time               `json:"updated_at"`
+	Result        string                  `json:"result"`
+	Steps         []RunStepDetails        `json:"steps"`
+	HasRfaResults bool                    `json:"has_wisp_results"`
```
jbarber: Did you rename the field in the API? Otherwise we're going to be stuck with this name forever and it's going to make no sense when we come to document it in the API.

We could just add an alias in the API?

kruszczynski: I wanted to do it later but I do see your point, let's do it now

PR #377, opened 2020-06-09
RF-18377 Enable GitHub's code scanning feature

jbarber:
Turn on code scanning to see how it works.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,52 @@
+name: "Code scanning - action"
+
+on:
+  push:
+  pull_request:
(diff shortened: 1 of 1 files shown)


kruszczynski:
It looks like `push` and `pull_request` trigger the checks twice?
Screenshot 2020-06-09 at 12 47 42

jbarber:
@kruszczynski yeah - I presume that's because of the:
```
on:
  push:
  pull_request:
  schedule:
    - cron: '0 10 * * 2'
```

The proposed config came from GH itself, so I presume it's ok to get going with - and I'm not concerned about the cost of GH actions at the moment.

PR #379, opened 2020-07-03
[RF-18165] Add rerun command

magni-:
We've had the endpoint for ages, but a wrapper for it was never added to the CLI (until now 🎉). The implementation and tests are heavily inspired by the "create a run from a run group" flow.

```bash
> ./rainforest-cli rerun 622054 --debug
Trying https://app.rainforestqa.com/api/1/runs/622054/rerun_failed...connected
POST HTTP/1.1
User Agent: rainforest-cli/2.15.3 [rainforest golang lib/2.0.0]
Host: app.rainforestqa.com

2020/07/03 21:59:32 Run 622146 has been created.
```
=> https://app.rainforestqa.com/runs/622146

This also adds `id` attributes to the `testsuite` and `testcase` nodes of the JUnit file, to make it real easy to get the created run ID when using the `--junit-file` option:
```bash
> ./rainforest-cli rerun 622054 --junit-file results.xml
...
> xmllint --xpath "string(testsuite/@id)" results.xml
622146
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ rainforest run --run-group <run_group_id>
 
 Rerun failed tests from a run.
 
+If you don't pass in an argument, it will try to get a run ID from the `RAINFOREST_RUN_ID` environment variable.
+
 ```bash
 rainforest rerun <failed_run_id>
 ```
--- a/README.md
+++ b/README.md
@@ -89,6 +89,12 @@ Run a run group.
 rainforest run --run-group <run_group_id>
 ```
 
+Rerun failed tests from a run.
+
+```bash
+rainforest rerun <failed_run_id>
+```
+
 #### Creating and Managing Tests
 
 Create new Rainforest test in RFML format (Rainforest Markup Language).
--- a/README.md
+++ b/README.md
@@ -81,6 +81,14 @@ Run individual tests in the foreground and report.
 rainforest run <test_id1> <test_id2>
 ```
 
+Run a run group.
+
+⚠️ This uses the configuration defined in the run group (environment, browsers, crowd, location). If you wish to run tests from a run group without using the run group's configuration, you will need to use the Rainforest API directly, passing a `run_group_id` parameter to [the `POST /runs` endpoint](https://app.rainforestqa.com/docs#!/runs/post1Runs). ⚠️
+
+```bash
+rainforest run --run-group <run_group_id>
+```
+
 #### Creating and Managing Tests
 
 Create new Rainforest test in RFML format (Rainforest Markup Language).
--- a/rainforest/runner.go
+++ b/rainforest/runner.go
@@ -76,20 +76,26 @@ func (c *Client) CreateRun(params RunParams) (*RunStatus, error) {
 }
 
 func validateRerunParams(params RunParams) error {
-	if params.Tags != nil {
-		return errors.New("Tags cannot be specified for rerun")
-	}
-	if params.Browsers != nil {
-		return errors.New("Browsers cannot be specified for rerun")
-	}
 	if params.Tests != nil {
 		return errors.New("Tests cannot be specified for rerun")
 	}
 	if params.RFMLIDs != nil {
 		return errors.New("RFML tests cannot be specified for rerun")
 	}
+	if params.Tags != nil {
+		return errors.New("Tags cannot be specified for rerun")
+	}
+	if params.SmartFolderID != 0 {
+		return errors.New("Folder cannot be specified for rerun")
+	}
+	if params.SiteID != 0 {
+		return errors.New("Site cannot be specified for rerun")
+	}
 	if params.Crowd != "" {
 		return errors.New("Crowd cannot be specified for rerun")
+	}
+	if params.Browsers != nil {
+		return errors.New("Browsers cannot be specified for rerun")
 	}
 	if params.Description != "" {
 		return errors.New("Description cannot be specified for rerun")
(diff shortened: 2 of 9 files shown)


runner.go
```diff
@@ -95,6 +101,34 @@ func (r *runner) startRun(c cliContext) error {
 		return nil
 	}
 
+	if c.Bool("se") {
+		os.Setenv("RAINFOREST_RUN_ID", strconv.Itoa(runStatus.ID))
```
jbarber, line 105: This won't do what you want, because the environment variable is set in the current process and child processes, but not in the parent.

I don't think it's possible to set environment variables in parent processes without using `ptrace()` or other shinangins. 

I tested this locally with the following program:
```
package main

import "os"

func main() {
        os.Setenv("RAINFOREST_RUN_ID", "1234")
}
```
and doing:
```
go run test.go; echo '$RAINFOREST_RUN_ID'
```
didn't print anything.

runner.go
```diff
@@ -375,6 +401,33 @@ func (r *runner) makeRunParams(c cliContext, localTests []*rainforest.RFTest) (r
...
+	if runIDString == "" {
+		return rainforest.RunParams{}, errors.New("Missing run ID")
+	}
+	runID, err = strconv.Atoi(runIDString)
+	if err != nil {
+		return rainforest.RunParams{}, errors.New("Invalid run ID specified")
+	}
+
+	var conflict string
+	if conflict = c.String("conflict"); conflict != "" && conflict != "abort" && conflict != "abort-all" {
```
jbarber, line 421: This validation logic is duplicated elsewhere, it could be moved to an `Action` block in the CLI parser?:
https://github.com/rainforestapp/rainforest-cli/blob/master/rainforest-cli.go#L193

runner.go
```diff
@@ -375,6 +401,33 @@ func (r *runner) makeRunParams(c cliContext, localTests []*rainforest.RFTest) (r
 	}, nil
 }
 
+func (r *runner) makeRerunParams(c cliContext) (rainforest.RunParams, error) {
+	var err error
+
+	var runID int
+	runIDString := c.Args().First()
+	if runIDString == "" {
+		runIDString = os.Getenv("RAINFOREST_RUN_ID")
```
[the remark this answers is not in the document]
magni-: That's for flags only unfortunately, arguments are more of a hassle. There's an [open issue](https://github.com/urfave/cli/issues/1074) about making arguments behave more like flags.

runner.go
magni-: `Action` blocks aren't for flags. There's no first-class support for enums yet, but there's a workaround shown in the [open issue](https://github.com/urfave/cli/issues/602).

rainforest/runner.go
```diff
@@ -68,6 +75,47 @@ func (c *Client) CreateRun(params RunParams) (*RunStatus, error) {
 	return &runStatus, nil
 }
 
+func validateRerunParams(params RunParams) error {
```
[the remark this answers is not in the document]
magni-: Reordered the params both here and in `validateRunGroupParams` (which is where the order I initially used came from).

rainforest/runner_test.go
```diff
@@ -51,6 +51,117 @@ func TestCreateRun(t *testing.T) {
...
+	mux.HandleFunc(fmt.Sprintf("/runs/%v/rerun_failed", runID), func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprint(w, `{"id": 123, "state":"in_progress"}`)
+	})
+	for _, params := range badParams {
+		_, err := client.CreateRun(params)
+		if err == nil {
+			t.Errorf("Expected error for params %v but there was no error", params)
+		}
+	}
+	cleanup()
```
grodowski, line 162: No need to call `cleanup()` here, since it's deferred in L56?
```suggestion
```

runner.go
grodowski: Are you planning to use this workaround? I guess the downside of it is the custom `EnumValue` struct def we'd have to add/maintain. IMO a possible approach could be to extract a simple `validateConflict(c cliContext) (string, error)` func to avoid the duplicate check

magni-: > IMO a possible approach could be to extract a simple validateConflict(c cliContext) (string, error) func to avoid the duplicate check

Planning on doing that and staying with the existing `StringFlag` for now.

rainforest/runner_test.go
magni-: Good catch, this was copy-pasted from `TestCreateRunFromRunGroup` 🙈 I'll remove it from both tests 👍

README.md
```diff
@@ -81,6 +81,22 @@ Run individual tests in the foreground and report.
 rainforest run <test_id1> <test_id2>
 ```
 
+Run a run group.
+
+⚠️ This uses the configuration defined in the run group (environment, browsers, crowd, location). If you wish to run tests from a run group without using the run group's configuration, you will need to use the Rainforest API directly, passing a `run_group_id` parameter to [the `POST /runs` endpoint](https://app.rainforestqa.com/docs#!/runs/post1Runs). ⚠️
```
roman-dowakin, line 86: `This uses the configuration defined in the run group: **crowd**`, let's keep the comment, but actually we have bug/feature here.
Even if a run group in web has Automation crowd, when it's run `rainforest-cli run --run-group id` it will use Testers actually.
To run with automation `--crowd` parameter is required

magni-: Yep we've run into this bug... but that's on the backend, not the CLI 👍

magni-: Looking into this some more... the bug _is_ on the CLI: https://github.com/rainforestapp/rainforest-cli/blob/05da067539ddce74f2f00729f7ccf7146b77795d/rainforest-cli.go#L188

PR #393, opened 2020-09-09
RF-19498 Compile static executable

roman-dowakin:
I'd like to release this to master and test if on `equinox` we'll get static executables


jbarber (approved):
Do we not need the arguments `-ldflags "-extldflags -static"`?

roman-dowakin:
> Do we not need the arguments `-ldflags "-extldflags -static"`?

It doesn't really matter for our case, so I don't want to add extra stuff.
Maybe it's more relevant to programs with C dependencies...

PR #397, opened 2020-12-14
Add the "rerun-if-last-failed" flag to the run command

maciejgryka:


rainforest/runner.go
```diff
@@ -156,3 +156,39 @@ func (c *Client) CheckRunStatus(runID int) (*RunStatus, error) {
...
 	return &runStatus, nil
 }
+
+func (c *Client) LastMatchingRun(params RunParams) (*RunStatus, error) {
+	var runStatuses []RunStatus
+	var lastMatchingRunStatus RunStatus
+	endpoint := "/runs?page_size=1"
+
+	// Get the last run with identical params
+	searchParams := fmt.Sprintf("&run_group_id=%v", params.RunGroupID)
```
maciejgryka, line 166: We still need to add other params here in addition to `run_group_id`.

rainforest-cli.go
```diff
@@ -246,6 +246,10 @@ func main() {
 					Name:  "wait, reattach",
 					Usage: "monitor existing run with `RUN_ID` instead of starting a new one.",
 				},
+				cli.BoolFlag{
```
magni-, line 249: What about using an `IntFlag` (`max-retries` ?) instead? That we'd default to `0`.

rainforest/runner.go
```diff
@@ -156,3 +156,39 @@ func (c *Client) CheckRunStatus(runID int) (*RunStatus, error) {
 
 	return &runStatus, nil
 }
+
+func (c *Client) LastMatchingRun(params RunParams) (*RunStatus, error) {
```
magni-, line 160: I don't think we need this? Instead in `monitorRunStatus` we can do something like (along with a `max-retries` flag)

```go
if done {
    if status.Result != "passed" {
        remainingRetries = c.Int("max-retries") 
        if remainingRetries > 0 {
            // set run ID for rerun
            os.Setenv("RAINFOREST_RUN_ID", strconv.Itoa(runID))
            // create new context with only decremented max-retries
            flagSet = cli.flag.NewFlagSet("rerun", 0)
            flagSet.Int("max-retries", remainingRetries - 1)
            newContext = cli.NewContext(nil, flagSet)
            return rerunRun(newContext)
        } else {
            return cli.NewExitError("", 1)
        }
}
```

Issue #396, opened 2020-12-11
Add retry limit parameter to cli to retry failed tests automatically for easier integration


BlankMatter:
### Problem
We are using Rainforest qa extensively and currently run into flaky tests which are based on the Rainforest qa virtual machine performance. In addition tests can also be flaky by nature if the application they are testing is not well designed. Using a service like gitlab with the Rainforest cli leads to the problem that when hitting the "retry" button on a failed job triggers all rainforest tests again, if failed or not. This leads to following issues:

- All rainforest tests of that run-group are triggered again, which makes it more expensive and the pipeline takes longer than if only the failed tests would be re-triggered
- Flaky tests that succeeded in the previous run may now fail since if every flaky test has a certain percentage to fail, re-running all tests does not decrease the overall chance for the pipeline to succeed
- Rainforest tests manually re-triggered in the app.rainforestqa console won't be reported back to the failed stage since connection is already terminated and the process exited with a failure code

### Proposed Solution
Adding an additional parameter to the cli run command that allows to define how often failed tests can be retried would allow to only re-run failed tests without the need of adding a custom script or using a workaround like chaining the retry command independent of if the first run succeeded or not. This is a common concept used in many ci services allowing the developer to set a maximum retry limit. The default limit if not parameter is given could be set to 1 which would not have any impact on current setups of the cli.

The parameter could be used like this
`docker run gcr.io/rf-public-images/rainforest-cli run --run-group 1234 --max-retry 3`

### Benefit
With only a single parameter the problem would be universally solved for any shell or environment. It also makes the rerun command independent of the RAINFOREST_RUN_ID variable in the environment

maciejgryka:
Thanks for the suggestion @BlankMatter, we're looking into it.

maciejgryka:
@BlankMatter after playing with it for a while, my favorite solution so far would be adding a flag to the `run` command, which would either start a new run OR re-run the last run with the same settings if the last one failed.

Specifically, you'd do something like `rainforest-cli run --run-group 123 --rerun-if-last-failed` and could run it in a loop until it succeeds. If it fails and you run it again, it'd find the last run with the same Run Group, realize it failed and re-run just the failed tests from that. If it fails again and you run the same command again, it'd look at the most recent re-run and re-run the failed test from that etc.

This still requires some adjustments to your CI, but they feel pretty easy (e.g. you could use [GitLab's retry](https://docs.gitlab.com/ee/ci/yaml/#retry)).

[...]

PR #398, opened 2020-12-16
[RF-20430] Add --max-reruns flag to the run command

maciejgryka:
This should address #396. When retrying, the console output looks like this:

```
2020/12/17 16:26:38 Run 674699 is in_progress and is 0% complete
2020/12/17 16:26:44 Run 674699 is in_progress and is 50% complete
2020/12/17 16:26:49 Run 674699 is in_progress and is 50% complete
2020/12/17 16:26:55 Run 674699 is in_progress and is 50% complete
2020/12/17 16:27:00 Run 674699 is in_progress and is 50% complete
2020/12/17 16:27:05 Run 674699 is in_progress and is 50% complete
2020/12/17 16:27:11 Run 674699 is in_progress and is 50% complete
2020/12/17 16:27:16 Run 674699 is now complete and has failed
2020/12/17 16:27:16 Creating JUnit report for run #674699: output.xml
2020/12/17 16:27:16 Fetching details for run #674699
2020/12/17 16:27:16 Fetching information for failed test #273949
2020/12/17 16:27:17 JUnit report successfully written to /home/mgryka/dev/rainforest/rainforest-cli/output.xml
2020/12/17 16:27:17 Retrying run 674699, attempt 1
2020/12/17 16:27:18 Run 674700 has been created.
2020/12/17 16:27:18 Run 674700 is queued and is 0% complete
2020/12/17 16:27:24 Run 674700 is in_progress and is 0% complete
2020/12/17 16:27:30 Run 674700 is in_progress and is 0% complete
2020/12/17 16:27:35 Run 674700 is in_progress and is 0% complete
2020/12/17 16:27:40 Run 674700 is in_progress and is 0% complete
2020/12/17 16:27:46 Run 674700 is in_progress and is 0% complete

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ Popular command line options are:
 - `--bg` - creates a run in the background and rainforest-cli exits immediately after. Do not use if you want rainforest-cli to track your run and exit with an error code upon run failure (ie: using Rainforest in your CI environment).
 - `--crowd [default|automation|automation_and_crowd|on_premise_crowd]` - select automation or your crowd of testers (for clients with on premise testers). For more information, contact us at help@rainforestqa.com.
 - `--wait RUN_ID` - wait for an existing run to finish instead of starting a new one, and exit with a non-0 code if the run fails. rainforest-cli will exit immediately if the run is already complete.
-- `--fail-fast` - fail the build as soon as the first failed result comes in. If you don't pass this it will wait until 100% of the run is done. Has no effect with `--bg`.
+- `--fail-fast` - return an error as soon as the first failed result comes in (the run always proceeds until completion, but the CLI will return an error code early). If you don't use it, it will wait until 100% of the run is done. Has no effect with `--bg` and cannot be used together with `--max-reruns`.
 - `--custom-url` - use a custom url for this run to use an ad-hoc QA environment on all tests. You will need to specify a `site_id` too for this to work. Note that we will be creating a new environment for your account for this particular run.
 - `--git-trigger` - only trigger a run when the last commit (for a git repo in the current working directory) has contains `@rainforest` and a list of one or more tags. E.g. "Fix checkout process. @rainforest #checkout" would trigger a run for everything tagged `checkout`. This over-rides `--tag` and any tests specified. If no `@rainforest` is detected it will exit 0.
 - `--description "CI automatic run"` - add an arbitrary description for the run.
@@ -343,6 +343,7 @@ api to construct a junit report.  This is useful to track tests in CI such as Je
 - `--import-variable-csv-file /path/to/csv/file.csv` - Use with `run` and `--import-variable-name` to upload new tabular variable values before your run to specify the path to your CSV file.
 - `--import-variable-name NAME` - Use with `run` and `--import-variable-csv-file` to upload new tabular variable values before your run to specify the name of your tabular variable. You may also use this with the `csv-upload` command to update your variable without starting a run.
 - `--single-use` - Use with `run` or `csv-upload` to flag your variable upload as `single-use`. See `--import-variable-csv-file` and `--import-variable-name` options as well.
+- `--max-reruns` - If set to a value > 0 and a test fails, the CLI will re-run failed tests a number of times before reporting failure. If `--junit-file <filename>` is also used, the JUnit reports of reruns will be saved under `<filename>.1`, `<filename>.2` etc. Cannot be used together with `--fail-fast`.
 
 ## Support
 
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ Popular command line options are:
 - `--run-group ID` - run/filter based on a run group. When used with `run`, this trigger a run from the run group; it can't be used in conjunction with other test filters.
 - `--environment-id` - run your tests using this environment. Otherwise it will use your default environment
 - `--conflict OPTION` - use the `abort` option to abort any runs in progress in the same environment as your new run. use the `abort-all` option to abort all runs in progress.
-- `--bg` - creates a run in the background and rainforest-cli exits immediately after. Do not use if you want rainforest-cli to track your run and exit with an error code upon run failure (ie: using Rainforest in your CI environment).
+- `--bg` - creates a run in the background and rainforest-cli exits immediately after. Do not use if you want rainforest-cli to track your run and exit with an error code upon run failure (ie: using Rainforest in your CI environment). Cannot be used together with `--max-reruns`.
 - `--crowd [default|automation|automation_and_crowd|on_premise_crowd]` - select automation or your crowd of testers (for clients with on premise testers). For more information, contact us at help@rainforestqa.com.
 - `--wait RUN_ID` - wait for an existing run to finish instead of starting a new one, and exit with a non-0 code if the run fails. rainforest-cli will exit immediately if the run is already complete.
 - `--fail-fast` - return an error as soon as the first failed result comes in (the run always proceeds until completion, but the CLI will return an error code early). If you don't use it, it will wait until 100% of the run is done. Has no effect with `--bg` and cannot be used together with `--max-reruns`.
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -54,9 +54,15 @@ var (
 // cli.Context, the other is fakeCLIContext which is used for testing.
 type cliContext interface {
 	String(flag string) (val string)
+	GlobalString(flag string) (val string)
 	StringSlice(flag string) (vals []string)
+	GlobalStringSlice(flag string) (vals []string)
 	Bool(flag string) (val bool)
+	GlobalBool(flag string) (val bool)
 	Int(flag string) (val int)
+	GlobalInt(flag string) (val int)
+	Uint(flag string) (val uint)
+	GlobalUint(flag string) (val uint)
 
 	Args() (args cli.Args)
 }
@@ -246,6 +252,10 @@ func main() {
 					Name:  "wait, reattach",
 					Usage: "monitor existing run with `RUN_ID` instead of starting a new one.",
 				},
+				cli.UintFlag{
+					Name:  "max-reruns",
+					Usage: "Rerun `max-reruns` times before reporting failure.",
+				},
 			},
 		},
 		{
@@ -276,6 +286,14 @@ func main() {
 				cli.StringFlag{
 					Name:  "junit-file",
 					Usage: "Create a JUnit XML report `FILE` with the specified name. Must be run in foreground mode.",
+				},
+				cli.UintFlag{
+					Name:  "max-reruns",
+					Usage: "Rerun `max-reruns` times before reporting failure.",
+				},
+				cli.UintFlag{
+					Name:  "rerun-attempt",
+					Usage: "Which rerun attempt this is.",
 				},
 			},
 		},
--- a/runner.go
+++ b/runner.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/rainforestapp/rainforest-cli/rainforest"
@@ -49,6 +50,17 @@ func (r *runner) startRun(c cliContext) error {
 			return cli.NewExitError(err.Error(), 1)
 		}
 		return monitorRunStatus(c, runID)
+	}
+
+	// Validate that --fail-fast and --max-reruns are not used together
+	failFast := c.Bool("fail-fast")
+	maxReruns := c.Uint("max-reruns")
+	if failFast && (maxReruns > 0) {
+		return cli.NewExitError(
+			"You can't use --fail-fast when --max-reruns is greater than 0. "+
+				"For the CLI to rerun on failure, it has to wait until completion.",
+			1,
+		)
 	}
 
 	var localTests []*rainforest.RFTest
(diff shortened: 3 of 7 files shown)


maciejgryka:
/reviewme @rainforestapp/devs 

This should address https://github.com/rainforestapp/rainforest-cli/issues/396.

runner.go
```diff
@@ -234,6 +243,12 @@ func filterExecuteTests(tests []*rainforest.RFTest, tags []string, forceExecute,
 	return result
 }
 
+func augmentJunitFileName(junitFile string, retryAttempt int) (string, error) {
+	pattern := regexp.MustCompile(`(\.[0-9])*\.xml`)
```
jbarber, line 247: This assumes we'll always retry in the single digits, so is broken for larger values.

I think it's broken for cases like `.0.1.2.3.4.5.xml` as well. Because `(foo)*` means match the thing inside the parenthesis zero or more times?
```suggestion
	pattern := regexp.MustCompile(`(\.[0-9]+)?\.xml$`)
```

TBH - I think it'd be cleaner to just append the retry attempt to the end of the filename and remove the constraint we've added about the file having to have an `.xml` suffix at 
https://github.com/rainforestapp/rainforest-cli/blob/2a2d3ad7576d097422c593d1bf5ed0723f16ca19/reporter.go#L93

This requirement is bad because:
1. maybe the user doesn't want to have a file ending in `.xml`
2. because the check is only made at the end of the run - so we could complete the run and then blow up and not save the junit output because the user didn't give a file name that ends in `.xml`!

rainforest-cli.go
```diff
@@ -277,6 +285,14 @@ func main() {
...
 					Usage: "Create a JUnit XML report `FILE` with the specified name. Must be run in foreground mode.",
 				},
+				cli.IntFlag{
+					Name:  "max-retries",
+					Usage: "re-try the run `max-retries` times before reporting failure.",
+				},
+				cli.IntFlag{
+					Name:  "retry-attempt",
+					Usage: "which attempt is this",
+				},
```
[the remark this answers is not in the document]
maciejgryka: Yeah, good catch, I'll change to `UintFlag`.

runner.go
```diff
@@ -242,6 +257,27 @@ func monitorRunStatus(c cliContext, runID int) error {
...
+					cmd, envv, _ := buildRerunArgs(c, runID)
+					exec_err := syscall.Exec("rainforest-cli", cmd, envv)
+					if exec_err != nil {
+						return cli.NewExitError(exec_err.Error(), 1)
+					}
+				} else {
+					return cli.NewExitError("", 1)
+				}
+			}
+
```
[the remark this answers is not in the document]
maciejgryka: Definitely, cleaning up

runner.go
maciejgryka: Good point, I'll get rid of the check and append `.1` to the end of the filename.

runner.go
```diff
@@ -242,6 +257,27 @@ func monitorRunStatus(c cliContext, runID int) error {
...
+					// At this point, we're either going to retry (by doing
+					// syscall.Exec below) or quit. Either way, we need to
+					// save the report first.
+					postRunJUnitReport(c, runID)
+				}
+
+				remainingRetries := c.Int("max-retries") - c.Int("retry-attempt")
+				if remainingRetries > 0 {
+					cmd, envv, _ := buildRerunArgs(c, runID)
+					exec_err := syscall.Exec("rainforest-cli", cmd, envv)
```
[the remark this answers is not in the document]
maciejgryka: I'll add a notification & put an example in the description.

runner.go
```diff
@@ -273,6 +309,47 @@ func monitorRunStatus(c cliContext, runID int) error {
...
+		strconv.Itoa(runID),
+		"--max-retries",
+		strconv.Itoa(maxRetries),
+		"--retry-attempt",
+		strconv.Itoa(retryAttempt + 1),
+	}
+	envv := []string{}
+
+	// Try to get the token from the env var and fall back to the CLI flag
+	if token := os.Getenv("RAINFOREST_API_TOKEN"); len(token) > 1 {
```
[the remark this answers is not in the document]
maciejgryka: I thought I tested it, but trying again, you're right. Removing the parts managing this env var.

maciejgryka:
/reviewme @jbarber 

Addressed the comments.

runner.go
```diff
@@ -51,6 +52,13 @@ func (r *runner) startRun(c cliContext) error {
 		return monitorRunStatus(c, runID)
 	}
 
+	// Validate that --fail-fast and --max-retries are not used together
+	failFast := c.Bool("fail-fast")
+	maxRetries := c.Uint("max-retries")
+	if failFast && (maxRetries > 0) {
```
jbarber, line 58: Hmm, this check does mean that the following invocation is invalid:
```
$ rainforest-cli --skip-update run --max-retries 1 --fail-fast 123
2020/12/17 15:43:34 The --fail-fast and --max-retries flags cannot be used together
```

but this is fine:
```
$ rainforest-cli --skip-update run --max-retries 1 --fail-fast 123
2020/12/17 15:43:54 RF API Error (404): Cannot find record
```

which is a bit weird?

runner.go
```diff
@@ -234,6 +242,10 @@ func filterExecuteTests(tests []*rainforest.RFTest, tags []string, forceExecute,
 	return result
 }
 
+func augmentJunitFileName(junitFile string, retryAttempt uint) string {
+	return fmt.Sprintf("%v.%v", junitFile, retryAttempt)
+}
```
jbarber, line 247: If we retry say three times, does this end up with files called:
* `foo.xml`
* `foo.xml.1`
* `foo.xml.1.2`

?

jbarber (commented):
Just think there's an issue with the adding of the suffix on the junit file.

README.md
```diff
@@ -343,6 +343,7 @@ api to construct a junit report.  This is useful to track tests in CI such as Je
 - `--import-variable-csv-file /path/to/csv/file.csv` - Use with `run` and `--import-variable-name` to upload new tabular variable values before your run to specify the path to your CSV file.
 - `--import-variable-name NAME` - Use with `run` and `--import-variable-csv-file` to upload new tabular variable values before your run to specify the name of your tabular variable. You may also use this with the `csv-upload` command to update your variable without starting a run.
 - `--single-use` - Use with `run` or `csv-upload` to flag your variable upload as `single-use`. See `--import-variable-csv-file` and `--import-variable-name` options as well.
+- `--max-retries` - If set to a value > 0 and a test fails, the CLI will re-run failed tests a number of times before reporting failure. Cannot be used together with `--fail-fast`.
```
magni-, line 346: Could you document what happens if you use this with the `--junit-file` parameter?

runner.go
magni-: Find/replace if `retryAttempt > 1` ?

runner.go
magni-: > This requirement is bad because:
> 
>     1. maybe the user doesn't want to have a file ending in `.xml`

Is that really a thing? I feel like the opposite is more likely. Will Circle for example recognize a JUnit file if it doesn't have the `xml` extension? I can't tell from [the docs](https://circleci.com/docs/2.0/collect-test-data/).
 
>     2. because the check is only made at the end of the run - so we could complete the run and then blow up and not save the junit output because the user didn't give a file name that ends in `.xml`!

That's a very good point, but we could just move the check to before the run gets kicked off?

jbarber: > Will Circle for example recognize a JUnit file if it doesn't have the xml extension

I don't think that's a concern of the CLIs and we have specific tooling (the orb) to handle running the CLI in the context of CircleCI. The norms for unix CLIs don't expect filenames to mean much - the only one I can think of that I normally use is `unzip`, and I normally curse that for being un-ergonomic for exactly this reason.

I also know I would curse a utility if it forced me to have a filename of a particular format when I wanted it in another, because then I have to write stuff to handle renaming it. It also means we can't specify `/dev/fd/3` as an output to allow users to pipe the output somewhere else (as long as the FD functions we're using don't do anything funky, this should just work TM). It's just a needless restriction IMO.

> but we could just move the check to before the run gets kicked off

Sure - still don't think the CLI should enforce filenames.

runner.go
maciejgryka: Wait, what's the difference? Did you mean to do `--max-retries 0` in the second one?

jbarber: Yep - bad copy/paste.

jbarber:
```
$ ./rainforest-cli --skip-update run --max-retries 0 --fail-fast 123
2020/12/18 10:15:54 RF API Error (404): Cannot find record
$
```

maciejgryka: Right, so what's weird about that? `max-retries` defaults to 0, so if you set it to `0` it's the same as not setting it, since there will be no reruns - so you can do `fail-fast`.

runner.go
maciejgryka: I'm guilty of regex.

runner.go
jbarber: It's weird because the error message says that `The --fail-fast and --max-retries flags cannot be used together` but it turns out they can - so ideally the error message should reflect this or the conditional should be "did the user set the fail-fast and max-retries argument" - I don't know if the library we're using lets us determine that though.

It's not worth sweating about too much.

maciejgryka: Updated the error message to
```
You can't use --fail-fast when --max-retries is greater than 0. For the CLI to retry on failure, it has to wait until completion.
```

runner.go
```diff
@@ -234,6 +243,11 @@ func filterExecuteTests(tests []*rainforest.RFTest, tags []string, forceExecute,
 	return result
 }
 
+func augmentJunitFileName(junitFile string, retryAttempt uint) string {
+	pattern := regexp.MustCompile(".xml[.0-9]*$")
+	return pattern.ReplaceAllString(junitFile, fmt.Sprintf(".xml.%v", retryAttempt))
+}
```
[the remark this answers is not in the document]
maciejgryka: You're right of course, this is so dumb; fixing.

runner.go
```diff
@@ -242,14 +256,25 @@ func monitorRunStatus(c cliContext, runID int) error {
...
-
 			postRunJUnitReport(c, runID)
 
 			if status.Result != "passed" {
-				return cli.NewExitError("", 1)
+				retryAttempt := c.Uint("retry-attempt")
+				remainingRetries := c.Uint("max-retries") - retryAttempt
+				if remainingRetries > 0 {
+					cmd, _ := buildRerunArgs(c, runID)
+					log.Printf("Retrying run %v, attempt %v", runID, retryAttempt+1)
```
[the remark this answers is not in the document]
maciejgryka: Also good point.

runner.go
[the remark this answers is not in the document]
maciejgryka: Fixed, I think - I moved `augmentJunitFileName` to `reporter`, which makes it all much simpler: we just pass the original `junit-file` value around each time and simply add `.<rerunAttempt>` without having to worry about building up `.1.2.3` etc.

jbarber (approved):
One code suggestion, otherwise :shipit:

README.md
```diff
@@ -330,7 +330,7 @@ Popular command line options are:
 - `--bg` - creates a run in the background and rainforest-cli exits immediately after. Do not use if you want rainforest-cli to track your run and exit with an error code upon run failure (ie: using Rainforest in your CI environment).
 - `--crowd [default|automation|automation_and_crowd|on_premise_crowd]` - select automation or your crowd of testers (for clients with on premise testers). For more information, contact us at help@rainforestqa.com.
 - `--wait RUN_ID` - wait for an existing run to finish instead of starting a new one, and exit with a non-0 code if the run fails. rainforest-cli will exit immediately if the run is already complete.
-- `--fail-fast` - fail the build as soon as the first failed result comes in. If you don't pass this it will wait until 100% of the run is done. Has no effect with `--bg`.
+- `--fail-fast` - return an error as soon as the first failed result comes in (the run always proceeds until completion, but the CLI will return an error code early). If you *don't use it, it will wait until 100% of the run is done. Has no effect with `--bg` and cannot be used together with `--max-reruns`.
```
magni-, line 333: Is that extra `*` a typo?
```suggestion
- `--fail-fast` - return an error as soon as the first failed result comes in (the run always proceeds until completion, but the CLI will return an error code early). If you don't use this flag, it will wait until 100% of the run is done. Has no effect with `--bg` and cannot be used together with `--max-reruns`.
```

rainforest-cli.go
```diff
@@ -13,7 +13,7 @@ import (
 
 const (
 	// Version of the app in SemVer
-	version = "2.17.0"
+	version = "2.18.0"
```
magni-, line 16: This should be changed in a followup [changelog PR](https://github.com/rainforestapp/rainforest-cli#changelog-pr).

runner.go
```diff
@@ -273,6 +296,34 @@ func monitorRunStatus(c cliContext, runID int) error {
...
+	}
+
+	if token := c.GlobalString("token"); len(token) > 0 {
+		cmd = append(cmd, "--token", token)
+	}
+	if conflict := c.String("conflict"); len(conflict) > 0 {
+		cmd = append(cmd, "--conflict", conflict)
+	}
+	if background := c.Bool("background"); background {
+		cmd = append(cmd, "--background")
```
magni-, line 318: This flag won't work with `--max-retries` for the same reasons as `--fail-fast` won't right? If you kick off a run with `--background`, the command will finish before the run does.

magni- (commented):
Sorry I missed the `--background` stuff on the last pass through 🤦‍♂️ Should be good to go once that's fixed up 🚀

runner.go
maciejgryka: Yes, good point, fixing.

magni- (approved):
Nitpick, but can you rebase to get rid of the version bumping/unbumping? Otherwise 🚀

PR #401, opened 2021-01-26
RF-20670 Locate the rainforest binary from the executable

jbarber:
Use `os.Executable()` (golang >=1.8) to get the path to the RF CLI when
re-execing it for reruns. This can fail if the file is moved, but it
doesn't seem to be possible to exec based on a FD reliably, so this is
probably good enough.

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -262,8 +262,13 @@ func monitorRunStatus(c cliContext, runID int) error {
 				remainingReruns := c.Uint("max-reruns") - rerunAttempt
 				if remainingReruns > 0 {
 					cmd, _ := buildRerunArgs(c, runID)
+					path, err := os.Executable()
+					if err != nil {
+						return cli.NewExitError(err.Error(), 1)
+					}
+
 					log.Printf("Rerunning %v, attempt %v", runID, rerunAttempt+1)
-					exec_err := syscall.Exec("rainforest-cli", cmd, []string{})
+					exec_err := syscall.Exec(path, cmd, []string{})
 					if exec_err != nil {
 						return cli.NewExitError(exec_err.Error(), 1)
 					}


jbarber:
Tested against RF with:
```
go build
mv rainforest-cli foo
RAINFOREST_API_TOKEN=.... ./foo run --skip-update --run-group 8495 --max-reruns=1
```

and checked that the binary was found for the rerun.

maciejgryka (approved):

BlankMatter:
Thank you for the quick fix Jonathan. As a suggestion the binary in the zip / tar file on the download page should be renamed rainforest-cli then if that is the general convention for the binary naming?

Here are the details on how to reproduce this:
 
The concrete issue is that following the instructions in the Readme the file gets unziped to a binary called "rainforest". According to Johnathan the binary needs to be called rainforest-cli otherwise it won't find itself to execute on a re-run.

This can be easily reproduced by:
1. Follow the download instructions in the readme file which leads to this page and following steps:
https://dl.equinox.io/rainforest_qa/rainforest-cli/stable
`
    curl -O https://bin.equinox.io/c/htRtQZagtfg/rainforest-cli-stable-darwin-amd64.zip
    unzip rainforest-cli-stable-darwin-amd64.zip -d /usr/local/bin
`
2. Unzip the binary and call the tests with the command `rainforest run <your env parameters> --max-reruns 3`

Result:
The test suit will run properly but when it fails it will only print an output of "path not found".

The generic message also made it hard to figure out what's actually going on, where I first assumed that the binary can't access the docker container after writing files for example.

jbarber:
@BlankMatter yep - I'll be looking at renaming things tomorrow, my concern with doing that is that it could break other people's build process if the binary is changed from rainforest => rainforest-cli and if they're grabbing `$LATEST`, so I want to make sure that I've considered the consequences before I do that.

jbarber:
Confirmed docker images works with this change:
```
$ docker images gcr.io/rf-public-images/rainforest-cli
REPOSITORY                              TAG     IMAGE ID      CREATED      SIZE
gcr.io/rf-public-images/rainforest-cli  latest  e9a24100b99f  3 hours ago  17.7 MB
$ docker run gcr.io/rf-public-images/rainforest-cli --version
Rainforest CLI version 2.18.1 - build: docker
$ docker run gcr.io/rf-public-images/rainforest-cli --skip-update  --token=94abf76cbde58d48336cae2358161957 run  --run-group 8495 --max-reruns=1
...
2021/01/27 14:15:08 Run 685803 is in_progress and is 50% complete
2021/01/27 14:15:13 Run 685803 is in_progress and is 50% complete
2021/01/27 14:15:24 Run 685803 is now complete and has failed   
2021/01/27 14:15:24 Rerunning 685803, attempt 1                 
2021/01/27 14:15:25 Run 685804 has been created.             
2021/01/27 14:15:26 Run 685804 is validating and is 0% complete
2021/01/27 14:15:31 Run 685804 is validating and is 0% complete
2021/01/27 14:15:37 Run 685804 is in_progress and is 0% complete
...
```

PR #402, opened 2021-01-27
RF-20682 Remove travis config

jbarber:
The build statuses from Travis don't appear to be getting back to
Github, and I'm not sure why we're building on both Travis and CircleCI.
Plus Travis are deprecating travis-ci.org and I don't want to do the
work to move over.

So remove Travis as a CI/CD platform.


jbarber:
Before merging this, remove the Travis webhook from the repo settings as well.

jbarber:
@ukd1 is there a good reason to keep building the CLI on both Circle and Travis? If not, I'd like to remove Travis from the build process.

magni- (approved):
🔥

PR #407, opened 2021-05-13
Adding run metadata / getting ready for CI specific output

ukd1:


gittrigger/gittrigger.go
```diff
@@ -35,12 +35,36 @@ func (g *gitTrigger) getLatestCommit() error {
...
+
+	if err != nil {
+		return "", err
+	}
+	remote_name := strings.TrimSpace(out.String())
+
+	out.Reset()
+	cmd.Stdout = &out
+
+	cmd = exec.Command("git", "remote", "get-url", "--all", remote_name)
```
ubergeek42, line 52: This worries me a bunch, since this could return a url like `http://user:pass@examplegitserver/repo.git`, and imo that's far too sensitive/risky.

gittrigger/gittrigger.go
```diff
@@ -35,12 +35,36 @@ func (g *gitTrigger) getLatestCommit() error {
 	return nil
 }
 
-func (g gitTrigger) checkTrigger() bool {
+func (g *gitTrigger) GetRemote() (string, error) {
+	var out bytes.Buffer
+	cmd := exec.Command("git", "remote")
```
ubergeek42, line 40: This can return multiple lines if you have more than one remote.

ubergeek42 (commented):
What is the use case for capturing the git remote?

gittrigger/gittrigger.go
ukd1: Great point; forgot some people don't use ssh keys - I'll parse it, strip any creds.

gittrigger/gittrigger.go
ukd1: Thanks, good catch

ukd1: Fixed in 1b02355

gittrigger/gittrigger.go
ukd1: Fixed in 1b02355

ukd1:
@ubergeek42 it's so that we can suggest / enable / develop better integrations - e.g. via linking up commits, or PRs with test results --> the correct repo. The CI info is for the same; suggesting a customer should use the Circle ORB, or equivalent.

rainforest/rainforest.go
```diff
@@ -113,7 +123,30 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
 
 	// Set UserAgent header with appended library version, will look like:
 	// "rainforest-cli/2.1.0 [rainforest golang lib/2.0.0]"
-	composedUserAgent := c.UserAgent + " [rainforest golang lib/" + libVersion + "]"
+	userAgent := []string{"rainforest", "golang", "lib/" + libVersion}
+
+	if c.SendTelemetry {
```
jbarber, line 128: Sending this info in the user agent is going to make parsing it harder over time, is there a reason to not send it via `X-` headers? 

This would also make this code simpler as we could just do:
```
if c.SendTelemetry {
  found, ci_name := detectci.WhichCI()
  if found {
    req.Header.Set('X-RF-CI-ENV', ci_name)
  }
  ...
}
```

gittrigger/gittrigger.go
```diff
@@ -14,18 +14,18 @@ type gitTrigger struct {
...
+	err := newGit.GetLatestCommit()
 	if err != nil {
 		return gitTrigger{}, err
 	}
 	return newGit, nil
 }
 
-func (g *gitTrigger) getLatestCommit() error {
-	cmd := exec.Command("git", "log", "-1", "--pretty=%B")
+func (g *gitTrigger) GetLatestCommit() error {
```
jbarber, line 26: Not sure if this should be exported as the result of it is already exposed via the `gitTrigger` struct as part of the object construction?

gittrigger/gittrigger.go
```diff
@@ -35,12 +35,36 @@ func (g *gitTrigger) getLatestCommit() error {
 	return nil
 }
 
-func (g gitTrigger) checkTrigger() bool {
+func (g *gitTrigger) GetRemote() (string, error) {
```
jbarber, line 38: Because the remotes aren't part of the `gitTrigger` struct this function isn't really part of the GitTrigger package? Not really a hard objection here because this module is confusing to me.

jbarber (commented):
Sending the CI env is fine and seems relatively uncontroversial to me.

The remotes stuff seems complicated and I'm not sure we can do anything useful with it ATM? Linking to branches/commits in e.g. GH would be cool, but to do that I think you'd need to have at least the commit SHA (and/or the branch info?), which isn't captured here. Normally (?) this info is exposed via an environment variable as well (e.g. `CIRCLE_PULL_REQUESTS`) - so getting it via that would probably be less risky than mucking about with `git` internals - plus it'd support non-git CVS.

ukd1:
> Sending the CI env is fine and seems relatively uncontroversial to me.
> 
> The remotes stuff seems complicated and I'm not sure we can do anything useful with it ATM? Linking to branches/commits in e.g. GH would be cool, but to do that I think you'd need to have at least the commit SHA (and/or the branch info?), which isn't captured here. Normally (?) this info is exposed via an environment variable as well (e.g. `CIRCLE_PULL_REQUESTS`) - so getting it via that would probably be less risky than mucking about with `git` internals - plus it'd support non-git CVS.

But, this would only work for CIRCLE - each CI works differently. Git is more popular than Circle. Also, if (and when) this fails, it doesn't crash, it just continues.

gittrigger/gittrigger.go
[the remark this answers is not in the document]
ukd1: I think this is wrong - I'm specifically asking for a single remote name; and you can't have multiple with the same name.

gittrigger/gittrigger.go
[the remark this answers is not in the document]
ukd1: I could, but I want to just get the current one, not all of them.

gittrigger/gittrigger.go
```diff
@@ -35,12 +35,36 @@ func (g *gitTrigger) getLatestCommit() error {
...
+	cmd.Stdout = &out
+
+	cmd = exec.Command("git", "remote", "get-url", "--all", remote_name)
+	cmd.Stdout = &out
+	err = cmd.Run()
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(strings.Split(strings.ReplaceAll(out.String(), "\r\n", "\n"), "\n")[0]), nil
```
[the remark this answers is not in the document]
ukd1: Well, this imho is actually not needed - I left it in just incase. The way I am doing it there will only ever be one returned; @ubergeek42 already commented the same, I wrote this + the tests and couldn't replicate two lines coming back.

rainforest/rainforest.go
ukd1: @jbarber as I want to get it done, and see if it works; if parsing it out via stack driver is too bad, I'll end up changing this to a header, or some other-way and storing it more explicitly - not just invthe log.

gittrigger/gittrigger.go
ukd1: @jbarber good spot; fixed by 77d6773da8b07082af6edf316daadde9e82e1aee

gittrigger/gittrigger.go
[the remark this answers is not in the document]
jbarber: The following is the situation I believe the code can't handle:
```
git init bar
cd bar
git remote add foo https://example-foo.com
git remote add moo https://example-moo.com
git remote get-url --all $(git remote)
```

This fails with:
```
usage: git remote get-url [--push] [--all] <name>

    --push                query push URLs rather than fetch URLs
    --all                 return all URLs
```
because `git remote` has multiple lines and my understanding of the `Trimspace` call on L47 is that it remotes leading + trailing spaces.

gittrigger/gittrigger.go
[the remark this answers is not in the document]
jbarber: My comment isn't about returning more than one origin's URL, it's whether we're returning the correct upstream origin for this branch.

rainforest/rainforest.go
jbarber: Ah, I thought you were going to do the parsing in the main app. If you've only got access to the User-Agent, then sticking it there makes sense.

ukd1: @jbarber yah - I figured getting this out, then using that first would be easiest. If it's crap, I'll move it.

gittrigger/gittrigger.go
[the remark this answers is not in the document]
ukd1: @jbarber ok, good catch - I was trying to add two remotes with the same name, not two remotes. Will fix.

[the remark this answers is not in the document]
ukd1: @jbarber fixed by 27bfb66ac3bc2aa7b9725b97e44658b1bab93e0c

gittrigger/gittrigger.go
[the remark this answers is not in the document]
ukd1: Addressed in 27bfb66ac3bc2aa7b9725b97e44658b1bab93e0c; still might not be the correct one - but most likely folks run one, and or use the same service for upstream. Making this perfect isn't the aim - getting enough data to work out where we need to integrate better is, and what we should support better.

gittrigger/gittrigger.go
```diff
@@ -35,12 +35,36 @@ func (g *gitTrigger) getLatestCommit() error {
 	return nil
 }
 
-func (g gitTrigger) checkTrigger() bool {
+func (g *gitTrigger) GetRemote() (string, error) {
+	var out bytes.Buffer
+	cmd := exec.Command("bash", "-c", "git remote | head -n 1")
```
ubergeek42, line 40: This probably isn't perfect, but good enough. Depending on the alphabetical ordering the first remote may not be the real remote. Terrible example to demonstrate:
```
$ git clone https://github.com/rainforestapp/rainforest-cli.git .
$ git remote add aaa http://example.org
$ git remote
aaa
origin
```

You'd need to do something like this SO question to be truly correct: https://stackoverflow.com/questions/12609415/find-out-which-remote-a-local-branch-is-tracking, but it gets tricky because it seems some of the commands are only supported with certain versions of git (though the question is old-ish so maybe we can just assume it'd always work)


Also, I don't think this needs to assume bash? Or avoid shell shenanigans entirely and get the last line using go directly (like you do further down to get the first line from the `git remote get-url` command).
```suggestion
	cmd := exec.Command("sh", "-c", "git remote | head -n 1")
```

jbarber, line 40: I don't think bash is available on OSX anymore by default, it's less likely to be present in containers as well. Won't work on Windows on command.com either (although I guess that's a niche use case). Stealing from your code below (and untested):

```suggestion
	cmd := exec.Command("git", "remote")
	cmd.Stdout = &out
	err := cmd.Run()

	if err != nil {
		return "", err
	}
	remote_name := strings.TrimSpace(strings.Split(strings.ReplaceAll(out.String(), "\r\n", "\n"), "\n")[0])
```

rainforest/rainforest.go
```diff
@@ -113,7 +123,30 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
...
+
+		var remote string
+		git, err := gitTrigger.NewGitTrigger()
+		if err == nil {
+			remote, err = git.GetRemote()
+			if err == nil {
+				u, err := giturls.Parse(remote)
+				if err == nil {
+					// Strip the user details, if any
+					u.User = nil
```
ubergeek42, line 142: It'd be nice if we had a test to explicitly make sure this is working properly (and avoid shipping a change which accidentally results in sending the user details later on)

ubergeek42 (approved):
Seems fine, a couple comments remain.

jbarber (approved):
Would suggest changing the `git remote | head` part to avoid shell where possible.

gittrigger/gittrigger.go
jbarber: Yeah, bash is very possibly not available in containers + new OSX releases. `|` will fail under `command.com` - although that's prolly niche for us.

jbarber:
> > Sending the CI env is fine and seems relatively uncontroversial to me.
> > The remotes stuff seems complicated and I'm not sure we can do anything useful with it ATM? Linking to branches/commits in e.g. GH would be cool, but to do that I think you'd need to have at least the commit SHA (and/or the branch info?), which isn't captured here. Normally (?) this info is exposed via an environment variable as well (e.g. `CIRCLE_PULL_REQUESTS`) - so getting it via that would probably be less risky than mucking about with `git` internals - plus it'd support non-git CVS.
> 
> But, this would only work for CIRCLE - each CI works differently. Git is more popular than Circle. Also, if (and when) this fails, it doesn't crash, it just continues.

Yeah I was worried about crashes on first review, but then saw you handled all the errors conditions.

Agreed git is more popular, can't really comment on the approach because it's not clear to me what the use case is.

PR #409, opened 2021-07-12
[RF-21913] Bypass Equinox update functionality after 2021-09-30

magni-:
Equinox is shutting down then: https://equinox.io/blog/shutting-down

We don't have any tests for `update.go`, I started going down the rabbit hole of writing some up and figuring out how to mock time in Go, but then figured just [using the playground](https://play.golang.org/p/fLfwgqKzTN1) would be fine given that we're about to rip this out anyways.

### DIFF
--- a/update.go
+++ b/update.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"log"
+	"time"
 
 	"github.com/equinox-io/equinox"
 	"github.com/urfave/cli"
@@ -19,6 +20,17 @@ eiFH+GffIMk09RFqkcMh5rIPu3ykm5V8
 `)
 
 func update(channel string, silent bool) error {
+	now := time.Now()
+	equinoxEOL, _ := time.Parse(time.RFC3339, "2021-09-30T00:00:00Z")
+	if now.After(equinoxEOL) {
+		if !silent {
+			log.Println("Equinox has shut down, update functionality has been disabled. " +
+						"To install the latest version, view our documentation: " +
+						"https://help.rainforestqa.com/docs/how-to-install-the-cli")
+		}
+		return nil
+	}
+
 	opts := equinox.Options{}
 	if channel == "" && releaseChannel != "" {
 		opts.Channel = releaseChannel


update.go
```diff
@@ -19,6 +20,15 @@ eiFH+GffIMk09RFqkcMh5rIPu3ykm5V8
 `)
 
 func update(channel string, silent bool) error {
+	now := time.Now()
+	equinoxEOL, _ := time.Parse(time.RFC3339, "2021-09-30T00:00:00Z")
+	if now.After(equinoxEOL) {
+		if !silent {
+			log.Println("Equinox has shut down, update functionality has been disabled")
```
ukd1, line 27: Maybe this would be a good place to link back here, and or to docs on how to update it themselves?

jbarber (approved):
Consider adding a message which appears now saying that the update functionality will be going away?

update.go
```diff
@@ -19,6 +20,17 @@ eiFH+GffIMk09RFqkcMh5rIPu3ykm5V8
 `)
 
 func update(channel string, silent bool) error {
+	now := time.Now()
+	equinoxEOL, _ := time.Parse(time.RFC3339, "2021-09-30T00:00:00Z")
```
jbarber, line 24: Please manually check this works with a time in the past, it'd suck to release this and have it blow up once we're past the `equinoxEOL`

magni-: I manually checked the parsing / time comparison code in the Go playground. Returning `nil` is copied from further down: https://github.com/rainforestapp/rainforest-cli/pull/409/files#diff-93ae3e89134c5b7d0b37869c5b950405c369d60559b300ed579b85b9f89f3884L40-L46

That said, I got a reply from Equinox overnight, looks like the plan is for their servers to respond with "no update available" responses (so as to not break existing clients), so there's not much urgency for this PR to go out.

magni-:
> Consider adding a message which appears now saying that the update functionality will be going away?

I shied away from adding that because it might not be? Looks like `goreleaser` and `go-github-selfupdate` should do the trick

magni-:
No longer needed after #413

PR #410, opened 2021-07-13
[RF-21915] Switch to GoReleaser to build and release

magni-:
I've left in the Equinox CI jobs for now, but removed references to it in the README.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,21 @@ jobs:
             CGO_ENABLED=0 equinox release --version=${CIRCLE_SHA1:0:8} --platforms="darwin_amd64 linux_amd64 darwin_386 linux_386 windows_amd64 windows_386" --signing-key=equinox.key --app=$EQUINOX_APP_ID --token=$EQUINOX_ACC_TOKEN --channel dev -- -ldflags "-X main.releaseChannel=dev -X main.build=${CIRCLE_SHA1:0:8}" github.com/rainforestapp/rainforest-cli
             rm -f equinox.key
 
+  release:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Release via GoReleaser
+          command: |
+            export GITHUB_TOKEN=$GITHUB_ACCESS_TOKEN
+            curl -sL https://git.io/goreleaser | bash
+      - run:
+          name: Show checksums
+          command: cat dist/checksums.txt
+      - store_artifacts:
+          path: dist/checksums.txt
+
 workflows:
   version: 2
   test_and_deploy:
@@ -97,3 +112,9 @@ workflows:
               only: /^v[0-9]+(\.[0-9]+)*$/
           context:
             - DockerHub
+      - release:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*(-.*)*/
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,31 @@
+# GoReleaser config documentation: https://goreleaser.com/customization/
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
+    - go mod tidy
+    # you may remove this if you don't need go generate
+    - go generate ./...
+builds:
+  - env:
+      - CGO_ENABLED=0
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -64,6 +64,21 @@ jobs:
...
             rm -f equinox.key
 
+  release:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Release via GoReleaser
+          command: |
+            export GITHUB_TOKEN=$GITHUB_ACCESS_TOKEN
```
magni-, line 74: We can't configure the name of the env var holding the GitHub token 😞 This line can be removed once the [project settings](https://app.circleci.com/settings/project/github/rainforestapp/rainforest-cli/environment-variables) are updated.

.circleci/config.yml
```diff
@@ -64,6 +64,21 @@ jobs:
...
+            curl -sL https://git.io/goreleaser | bash
```
jbarber, line 75: This seems terrible, but I see it's their recommend way of doing things...
https://goreleaser.com/ci/circle/

They have a docker image, would it be better to use that? If only so Circle might have a cached version of the image?

.goreleaser.yml
```diff
@@ -0,0 +1,38 @@
+# This is an example .goreleaser.yml file with some sane defaults.
```
jbarber, line 1: Hopefully it's not so much of an example anymore :grin:  
```suggestion
```

README.md
```diff
@@ -358,34 +358,25 @@ Email [help@rainforestqa.com](mailto:help@rainforestqa.com) if you're having tro
...
 1. Branch from master to update `CHANGELOG.md` to include the commit hashes and release date
 1. Update the `version` constant in `rainforest-cli.go` following [semantic versioning](http://semver.org/)
 1. Merge to master
-#### Release
-1. Release via [GitHub Releases](https://github.com/rainforestapp/rainforest-cli/releases)
-1. Update https://github.com/rainforestapp/homebrew-public to use the latest stable build url from [equinox](https://equinox.io)
+### Release
+1. Tag `master` after merging: `git tag vX.Y.Z && git push --tags`
+1. This will create a [draft GitHub Release](https://github.com/rainforestapp/rainforest-cli/releases). Edit the description as appropriate and publish the release.
+1. Update https://github.com/rainforestapp/homebrew-public to use the latest URL and SHA256. Both can be found in the GitHub Release assets. Additionally, the SHA256 is output as part of the CircleCI `Release` job.
```
jbarber, line 375: Followup PR, but looks like we can automate this? https://goreleaser.com/customization/homebrew/

README.md
```diff
@@ -358,34 +358,25 @@ Email [help@rainforestqa.com](mailto:help@rainforestqa.com) if you're having tro
...
-
-#### Development PR
+### Development PR
 1. Branch from master
 1. Do work
 1. Open PR against master
 1. Merge to master
-#### Changelog PR
+### Changelog PR
 1. Branch from master to update `CHANGELOG.md` to include the commit hashes and release date
```
jbarber, line 369: Orthogonal to this change, but wondering if we should drop the requirement for a sha in the changelog, as it would let us release with one less PR?

jbarber (approved):
YOLO

.circleci/config.yml
magni-: Their Docker image doesn't use the same version of Go as we do (it's on the latest 1.16 while we're still on 1.12). It _should_ be fine since all Go versions have been backwards compatible (and really we should just update this repo to use 1.16), but I wanted to minimize changes in this PR.

README.md
magni-: We definitely can. I looked into both this config option as well as how [GitHub](https://github.com/cli/cli/blob/trunk/.github/workflows/releases.yml#L169-L175) and [CircleCI](https://github.com/CircleCI-Public/circleci-cli/blob/master/.circleci/brew-deploy.sh) do it.

README.md
magni-: Maybe? It's nice to be able to easily group multiple smaller PRs into one update, and not that much of a hassle to open a separate release PR. But either way I'd rather not have that change in this PR.

PR #412, opened 2021-07-21
RF-22011 report urls for failures

jbarber:
Provide URLs to the app so it's easier to hop from the CircleCI display of failed results to the failure and diagnose what happened.

This generates junit XML that looks like:
```
<testsuite id="761688" name="Rainforest - master 188298 2021-07-20T13:50:15Z" tests="1" errors="0" failures="1" time="734.472">
  <testcase id="204138" name="Embed Test" time="580">
    <failure type="chrome_1440_900" message="https://app.rainforestqa.com/runs/761688/tests/204138">/failure>
  </testcase>
</testsuite>
```

which is rendered by Circle like this:
image

You can then hightlight the URL, right-click and select "go to ..." to open the failed test in a new tab. It'd be nicer if Circle gave you an actual clickable link - but I don't know how to get that out of their renderer.

### DIFF
--- a/reporter.go
+++ b/reporter.go
@@ -179,7 +179,7 @@ func testURL(runID int, testID int) string {
 	return u.String()
 }
 
-func createJUnitReportSchema(runDetails *rainforest.RunDetails, api reporterAPI) (*jUnitReportSchema, error) {
+func createJUnitReportSchema(runDetails *rainforest.RunDetails, apiReporter reporterAPI) (*jUnitReportSchema, error) {
 	type processedTestCase struct {
 		TestCase jUnitTestReportSchema
 		Error    error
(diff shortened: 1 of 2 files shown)


reporter.go
```diff
@@ -195,7 +202,7 @@ func createJUnitReportSchema(runDetails *rainforest.RunDetails, api reporterAPI)
 
 			if test.Result == "failed" {
 				log.Printf("Fetching information for failed test #" + strconv.Itoa(test.ID))
-				testDetails, err := api.GetRunTestDetails(runDetails.ID, test.ID)
```
jbarber, line 198: I found this super confusing - as it clobbers the globally available `api` object. So I renamed it

reporter_test.go
```diff
@@ -240,9 +241,9 @@ func TestCreateJUnitReportSchema(t *testing.T) {
 	}
 
 	// Dummy API - should not be used when there are no failed tests
-	api := newFakeReporterAPI(-1, []rainforest.RunTestDetails{})
```
jbarber, line 243: Again, don't clobber `api`. This change probably isn't required, but seems like a good idea.

ubergeek42 (approved):
👍  not clobbering globals!

The actual change looks pretty simple and makes sense, so I say :shipit:

jbarber:
Hey @ChristianBlank - after our conversation yesterday I whipped this up change to include the URLs to the app in the junit output for the CLI.

Before releasing it I wanted to confirm that this behaves in the same way as the implementation you scripted, and to give you a heads up this change is coming in case it breaks your scripts (and potentially your build process).

PR #416, opened 2021-08-11
[RF-21914] Stop publishing releases to Equinox

magni-:
The plan is to release this a `2.21.1`, and then we'll be able to tell anyone still using `2.20.0` and below that they need to update.


jaytennier (approved):
🔥

PR #418, opened 2021-08-30
[RF 22271] Add is_temporary flag when we create a temporary environment

ukd1:


jbarber (approved):

rainforest/environments.go
```diff
@@ -2,22 +2,25 @@ package rainforest
...
+	IsTemporary bool   `json:"is_temporary"`
 }
 
 // Environment represents an environment in Rainforest
 type Environment struct {
-	ID   int    `json:"id"`
-	Name string `json:"name"`
+	ID          int    `json:"id"`
+	Name        string `json:"name"`
+	IsTemporary bool   `json:"is_temporary"`
```
jbarber, line 14: Do we use the pattern of `is_XYZ` in the API already, or do we tend to just use `XYZ`?

ukd1: We do:

image

PR #420, opened 2021-09-02
Make it more obvious that you need to wait for the build to finish

magni-:
...before you can publish it.


ubergeek42 (approved):

ukd1:
@magni- thanks, this tripped me up being impatient!

PR #421, opened 2021-09-07
[RF-22338] Show a little more info in runs

ukd1:
Show a little more info in runs, and also better links to Rainforest's app when one starts.

https://rainforest.atlassian.net/browse/RF-22338

/reviewme @jbarber

### DIFF
--- a/rainforest/runner.go
+++ b/rainforest/runner.go
@@ -39,6 +39,8 @@ type RunStatus struct {
 		Total    int `json:"total"`
 		Complete int `json:"complete"`
 		NoResult int `json:"no_result"`
+		Passed   int `json:"passed"`
+		Failed   int `json:"failed"`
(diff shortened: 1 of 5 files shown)


rainforest/runner.go
```diff
@@ -39,6 +39,8 @@ type RunStatus struct {
 		Total    int `json:"total"`
 		Complete int `json:"complete"`
 		NoResult int `json:"no_result"`
+		Passed   int `json:"passed"`
+		Failed   int `json:"failed"`
```
ukd1, line 43: These two are passed by default, even if there are no tests with that state - https://github.com/rainforestapp/Rainforest/blob/85da6c4424376b7a85d72106c3d98d386a3707c3/app/models/run.rb#L608

jbarber (approved):
:tada:

PR #424, opened 2021-09-17
[RF 22358] Pull JUnit files from the API, instead of generating locally

ukd1:
/reviewme @magni- @maciejgryka


runner.go
```diff
@@ -259,7 +259,7 @@ func monitorRunStatus(c cliContext, runID int) error {
 		log.Print(msg)
 
 		if done {
-			postRunJUnitReport(c, runID)
+			// postRunJUnitReport(c, runID)
```
ukd1, line 262: This is WIP still, fyi 🤞

go.sum
```diff
@@ -1,5 +1,6 @@
 github.com/aws/aws-sdk-go v1.34.18 h1:Mo/Clq3u1dQFzpg8YQqBii8m+Vl3fWIfHi6kXs5wpuM=
 github.com/aws/aws-sdk-go v1.34.18/go.mod h1:5zCpMtNQVjRREroY7sYe8lOMRSxkhG6MZveU8YkpAk0=
+github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
```
ukd1, line 3: Locally, I can't run this without -- but not sure why it's not happened to anyone else?

resources.go
```diff
@@ -146,3 +150,70 @@ func printRunGroups(api resourceAPI) error {
...
+		if err != nil {
+			return cli.NewExitError(err.Error(), 1)
+		}
+	} else if deprecatedRunIDArg := c.String("run-id"); deprecatedRunIDArg != "" {
+		runID, err = strconv.Atoi(deprecatedRunIDArg)
+		if err != nil {
+			return cli.NewExitError(err.Error(), 1)
+		}
+
+		log.Println("Warning: `run-id` flag is deprecated. Please provide Run ID as an argument.")
```
magni-, line 189: When are we going to finally drop this 😅 This warning was originally added back in [November 2016](https://github.com/rainforestapp/rainforest-cli/commit/a7467957ca55402ae038d5ba375cad343288a5d3) (meaning it's been deprecated since we released `v2.0.0` the following spring).

ukd1: Ok, I'll drop it in this release 😇

.circleci/config.yml
```diff
@@ -1,11 +1,7 @@
-version: 2
-
+version: 2.1
 defaults: &defaults
   docker:
-    - image: circleci/golang:1.12.4-node
-      auth:
-        username: $DOCKERHUB_USERNAME
-        password: $DOCKERHUB_TOKEN
+    - image: cimg/go:1.17.1-node
```
ukd1, line 4: https://circleci.com/developer/images/image/cimg/go - it's the new replacement for the legacy one we were using

resources.go
```diff
@@ -146,3 +150,83 @@ func printRunGroups(api resourceAPI) error {
...
+	if fileName == "" {
+		return nil
+	}
+	api = rainforest.NewClient(c.String("token"), c.Bool("debug"))
+
+	cmd := []string{
+		"rainforest-cli",
+		"report", strconv.Itoa(runID),
+		"--skip-update", // skip auto-updates for reports inside a run
+	}
```
magni-, line 166: Why create and execute a new command rather than just calling `resource.GetRunJunit` here?

ukd1: I stated that way, but it needed more stuff brought in here to do than was nice. When I was fixing the reruns code's env, I thought that way kept it simpler.

magni- (approved):
🍪 points for cleaning up git history a bit before merging 🙇🏼

PR #425, opened 2021-09-21
RF-22483 - Publish a chocolatey package

ukd1:


ukd1:
Obviously this needs a squash....but the package is out and being reviewed. 
image

I'd love a review on this PR please. Ideas, but not strictly necessary? @maciejgryka thoughts?

- make a custom readme for the windows version, as the one in the package is just the first 4k of it ... for undocumented reasons; see chocolatey/choco#2007 🤷‍♂️
- run only on tags, not each time it's pushed - or manually and add to the docs. it just fails when it's pushed twice with the same version
- figure out and document rollback/pulling the package? 
- mutli-arch for chocolatey - nothing obvious here, it seems you just make more than one package 😵‍💫 

/reviewme @jbarber @maciejgryka

ukd1:
FYI, I tested this manually on Windows 10 -- building the package, and installing it, then running the binary. Works fine.

However, I can't / haven't tested the package from scratch -- as the package isn't published yet. Assuming it'll work, as it does when installed directly, but will check when it's released.

gep13:
> @ukd1 said...
However, I can't / haven't tested the package from scratch -- as the package isn't published yet. 

While in the moderation phase, you can install the package by directly requesting the specific version number, i.e:

```
choco install rainforest-cli --version=2.22.2
```

> @ukd1 said...
it just fails when it's pushed twice with the same version

This is very much by design. Once approved, a package is immutable, meaning that it can't be changed. A unique version number will be required for each package version that you want to publish to community.chocolatey.org.

> @ukd1 said...
mutli-arch for chocolatey - nothing obvious here, it seems you just make more than one package

This is definitely an area that we need to improve on, but there is absolutely no requirement to have multiple packages for different system architectures.  Looking at the package that is currently in moderation, you are using what we refer to as an embedded package.  To support different architectures, you would create a folder structure like the following:

* tools
  * x86
    * rainforest-cli.exe
  * x64
    * rainforest64.cli.exe

And from there, Chocolatey will do the "right" thing by shimming the current exe based on the architecture of the machine that is currently being installed onto. **NOTE:** I have made a guess here about the name of the 64 bit assembly, but it can be anything, even the same name.

Again, looking at the package in the moderation queue, it will remain in the current state until some required changes are made to the package:

> Requirements represent the minimum quality of a package that is acceptable. When a package version has failed requirements, the package version requires fixing and/or response by the maintainer. Provided a Requirement has flagged correctly, it must be fixed before the package version can be approved. The exact same version should be uploaded during moderation review.

> * The PackageSourceUrl element in the nuspec file should be a valid Url. Please correct this More...
> * Binary files (.exe, .msi, .zip, etc) have been included without including a LICENSE.txt file. This file is required when including binaries More...
> * Binary files (.exe, .msi, .zip) have been included without including a VERIFICATION.txt file. This file is required when including binaries More...

Once these changes are made, you should be able to push the same package version again.  **NOTE:** This is different to what I said above regarding immutability, and this is the _only_ exception to this rule.  While the package is being worked on, it can be changed, but once approved, a package version is immutable.

jbarber:
Thankyou for the feedback @gep13 ! Building a chocolatey package is new to us, so your comments are very useful in helping us to do the right thing.

chocolatey/main.rb
```diff
@@ -0,0 +1,121 @@
...
+# write the nuget
+builder = Builder::XmlMarkup.new(indent: 2)
+builder.instruct!(:xml, version: '1.0', encoding: 'UTF-8')
+
+xml = builder.package(xmlns: 'http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd') do |package|
+  package.metadata do |metadata|
+    metadata.title('rainforest-cli')
+    metadata.id('rainforest-cli')
+    metadata.version(latest_release.version)
+    metadata.summary('A command line interface to interact with Rainforest QA - https://www.rainforestqa.com/.')
```
jbarber, line 97:
```suggestion
    metadata.summary('A command line interface to interact with Rainforest QA - https://www.rainforestqa.com')
```

jbarber (approved):
The mechanics of it looks good. I think putting it in a separate repo is a good idea.

ukd1:
> > @ukd1 said...
> > However, I can't / haven't tested the package from scratch -- as the package isn't published yet.
> 
> While in the moderation phase, you can install the package by directly requesting the specific version number, i.e:
> 
> ```
> choco install rainforest-cli --version=2.22.2
> ```

@gep13 awesome, I'll add this to our testing so that we know it'll install itself post a push.

>  > @ukd1 said...
> > it just fails when it's pushed twice with the same version
> 
> This is very much by design. Once approved, a package is immutable, meaning that it can't be changed. A unique version number will be required for each package version that you want to publish to community.chocolatey.org.

@gep13 this was more meant as a note to the reviewers; i.e. I didn't make it our release process break, I expected y'all to handle duplicates like this.

> > @ukd1 said...
> > mutli-arch for chocolatey - nothing obvious here, it seems you just make more than one package
> 
> This is definitely an area that we need to improve on, but there is absolutely no requirement to have multiple packages for different system architectures. Looking at the package that is currently in moderation, you are using what we refer to as an embedded package. To support different architectures, you would create a folder structure like the following:
> 
> * tools
>   
>   * x86
>     
>     * rainforest-cli.exe
>   * x64
>     
>     * rainforest64.cli.exe
> 
> And from there, Chocolatey will do the "right" thing by shimming the current exe based on the architecture of the machine that is currently being installed onto. **NOTE:** I have made a guess here about the name of the 64 bit assembly, but it can be anything, even the same name.

Roger that, will check this out when I get a second. Thanks @gep13!

> Again, looking at the package in the moderation queue, it will remain in the current state until some required changes are made to the package:

Understood, this was our first pass.

> > Requirements represent the minimum quality of a package that is acceptable. When a package version has failed requirements, the package version requires fixing and/or response by the maintainer. Provided a Requirement has flagged correctly, it must be fixed before the package version can be approved. The exact same version should be uploaded during moderation review.
> 
> > * The PackageSourceUrl element in the nuspec file should be a valid Url. Please correct this More...
> > * Binary files (.exe, .msi, .zip, etc) have been included without including a LICENSE.txt file. This file is required when including binaries More...
> > * Binary files (.exe, .msi, .zip) have been included without including a VERIFICATION.txt file. This file is required when including binaries More...
> 
> Once these changes are made, you should be able to push the same package version again. **NOTE:** This is different to what I said above regarding immutability, and this is the _only_ exception to this rule. While the package is being worked on, it can be changed, but once approved, a package version is immutable.

Got it, thanks; I'll work on these updates shortly.

Again, many thanks for the help and jumping in here with your comments - very much appreciated! 👍🏻 💯

gep13:
@ukd1 you are very welcome! Feel free to reach out if you have any other questions. We know that we need to improve the docs in the area of package creation, as they are weak in this area, so feedback on stumbling blocks really helps.

ukd1:
@gep13 I'm not sure if they're weak, just complex as there are so many possibilities. Def would be easier with an end-to-end example for something simple (i.e. single standalone binary) -- though, maybe I missed it.

Also, like the 4k char limit, the other issues automatically raised would be amazing to know just from something like `choco validate` or something similar.  Even just outputting `https://community.chocolatey.org/packages/rainforest-cli/2.22.2` as soon as it was pushed would be amazing; that has most of the info in I need to fix things (along with the email).

FYI, will be closing this PR and moving all work to https://github.com/rainforestapp/rainforest-cli-chocolatey

gep13:
>@ukd1 said...
though, maybe I missed it.

No, unfortunately not, this is something that is missing, and something which has been on my todo list for a while 😢 We have some ideas around extending the `choco new` command, to introduce a wizard style approach, to prompt the user what type of package they are creating, and then tailor the templated files towards their selection.

> @ukd1 said...
from something like choco validate or something similar.

This is something that is on the roadmap. The plan is to take the package validation rules from the automated service, and bring them down into choco.exe. These would either be executed when you run `choco pack` or as a standalone command as you suggest, or perhaps both.  The rules would be the same as the following:

https://docs.chocolatey.org/en-us/community-repository/moderation/package-validator/

> @ukd1 said...
Even just outputting https://community.chocolatey.org/packages/rainforest-cli/2.22.2

Oooo, I like this idea. I will have to check what can be done here.

PR #434, opened 2021-09-30
Use rainforest rather than rainforest-cli in Docker image as well

magni-:
Keeps things consistent whether you're installing from GH releases or using our Docker image.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,12 +26,37 @@ commands:
             go version
   go_build:
     description: "go build, and output cli version"
+    parameters:
+      windows:
+        description: Whether this is running on Windows
+        type: boolean
+        default: false
     steps:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -30,8 +30,8 @@ commands:
       - run:
           name: Build
           command: |
-            go build
-            ./rainforest-cli -v --skip-update
+            go build -o rainforest
+            ./rainforest -v --skip-update
```
AJFunk, line 34: do we still want these `--skip-update`s?

magni-: I think so? The changes in this file are just so we build the same way in Circle as we do with GoReleaser and in the Dockerfile

ukd1 (approved):

PR #439, opened 2021-11-04
Fix docker binary names

ukd1:
/reviewme @ubergeek42


ukd1:
closed, as it's got more history in than it should -- my bad!

PR #442, opened 2021-12-03
[RF-23055] Add `--save-run-id` flag to `run` and `rerun` commands

magni-:
This is so that the CircleCI Orb and GitHub Actions can easily fetch the run ID for rerun purposes, without needing to install `libxml` to parse the JUnit each time (the workaround they're currently using).

I didn't add it to the README because I figured it's not of much use for other use cases? The flag is documented in the CLI itself.

I also updated `urfave/cli` to v1 latest since we were still on a super old SHA (not even a tag), which was annoying when reading their documentation. Upgrading to v2 involves breaking changes (most notably, it'll require users to set all flags before passing in arguments), so I opted against doing that in this PR.

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -108,6 +108,7 @@ func (r *runner) startRun(c cliContext) error {
 		return cli.NewExitError(err.Error(), 1)
 	}
 	r.showRunCreated(runStatus)
+	writeRunID(c, runStatus)
 
 	// if background flag is enabled we'll skip monitoring run status
 	if c.Bool("bg") {
@@ -128,6 +129,7 @@ func (r *runner) rerunRun(c cliContext) error {
 		return cli.NewExitError(err.Error(), 1)
 	}
 	r.showRunCreated(runStatus)
+	writeRunID(c, runStatus)
 
 	// if background flag is enabled we'll skip monitoring run status
 	if c.Bool("bg") {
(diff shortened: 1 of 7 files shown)


ukd1 (changes requested):
LGTM, but no tests?

runner.go
```diff
@@ -538,3 +540,22 @@ func expandStringSlice(slice []string) []string {
...
 }
+
+// writeRunID writes the run ID to a file if the flag is set
+func writeRunID(c cliContext, runStatus *rainforest.RunStatus) error {
+	filePath := c.String("save-run-id")
+	if filePath == "" {
+		return nil
+	}
+
+	file, err := os.Create(filePath)
```
jbarber, line 551: Does `os.Create` work with `filePath` having the value `/dev/stderr` / `/dev/stdout`? Not a big deal if it doesn't, but it'd be nice if the CLI accepts these special files.

runner.go
```diff
@@ -538,3 +540,22 @@ func expandStringSlice(slice []string) []string {
...
+	defer file.Close()
+
+	if err != nil {
+		return cli.NewExitError(err.Error(), 1)
+	} else {
+		file.WriteString(strconv.Itoa(runStatus.ID))
```
jbarber, line 557: Is this new line terminated? IMO it should be.

jbarber (approved):

runner.go
magni-: It is not because the files we currently generate in the Orb/Action aren't, but from testing, a newline shouldn't cause isses, so I'll add one 👍🏼

runner.go
magni-: Weirdly enough, it works... _once_. And then it completely breaks the `--save-run-id` flag and I need to recompile.

Ah, it turns out that happens regardless of what I pass in, not just for special values like `/dev/stdout` 🤔

magni-: Just running `./build/rainforest run --help` twice results in `--save-run-id` disappearing for the second execution 🤔

magni-:
@ukd1 added a test (after rewriting the whole config file to make it slightly less full of copy-paste?)
@jbarber added a newline to the run ID file

ukd1 (approved):

PR #445, opened 2022-01-12
RF-23061 Add --max-retries flag to the run command

maciejgryka:


ukd1 (approved):
LGTM - two things;

1. plz update the CHANGELOG
2. can you test what happens if you run with an invalid (to the api) value - i.e. 4 - does it error nicely? if not, I think it's worth fixing
3. naming of the flag imho `max-retries` infers it will work for all tests, and there is nothing (here) saying it won't. Please change to `automation-max-retries`, or otherwise make it really clear. I think the rename is best as it matches the API, and internals.

README.md
```diff
@@ -359,7 +359,8 @@ api to construct a junit report.  This is useful to track tests in CI such as Je
...
-- `--max-reruns` - If set to a value > 0 and a test fails, the CLI will re-run failed tests a number of times before reporting failure. If `--junit-file <filename>` is also used, the JUnit reports of reruns will be saved under `<filename>.1`, `<filename>.2` etc. Cannot be used together with `--fail-fast`.
+- `--max-reruns` - If set to a value > 0 and one or more tests fail, the CLI will create a new run with the failed tests a number of times before reporting failure. If `--junit-file <filename>` is also used, the JUnit reports of reruns will be saved under `<filename>.1`, `<filename>.2` etc. Cannot be used together with `--fail-fast`.
+- `--automation-max-retries` - If set to a value > 0 and a test fails, it will be retried within the same run, up to than number of times. If all retries fail, we report failure, if there is a pass, we report that and stop retrying. The failed-then-passed attempts do not affect the final result of the run, but can be inspected in the interface. See [our docs](https://help.rainforestqa.com/docs/test-retries) for more detail.
```
jbarber, line 363:
```suggestion
- `--automation-max-retries` - If set to a value > 0 and a test fails, it will be retried within the same run, up to that number of times. If all retries fail, we report failure, if there is a pass, we report that and stop retrying. The failed-then-passed attempts do not affect the final result of the run, but can be inspected in the web interface. See [our docs](https://help.rainforestqa.com/docs/test-retries) for more detail.
```

jbarber (approved):
What happens if you pass in a negative number?

maciejgryka:
@jbarber negative numbers are handled reasonably, since the flag value is defined as uint:

image

Similar when passing in value > 3:
image

Both could be better, but they seem good enough, IMO.

README.md
```diff
@@ -359,7 +359,8 @@ api to construct a junit report.  This is useful to track tests in CI such as Je
...
-- `--max-reruns` - If set to a value > 0 and a test fails, the CLI will re-run failed tests a number of times before reporting failure. If `--junit-file <filename>` is also used, the JUnit reports of reruns will be saved under `<filename>.1`, `<filename>.2` etc. Cannot be used together with `--fail-fast`.
+- `--max-reruns` - If set to a value > 0 and one or more tests fail, the CLI will create a new run with the failed tests a number of times before reporting failure. If `--junit-file <filename>` is also used, the JUnit reports of reruns will be saved under `<filename>.1`, `<filename>.2` etc. Cannot be used together with `--fail-fast`.
+- `--automation-max-retries` - If set to a value > 0 and a test fails, it will be retried within the same run, up to that number of times. If all retries fail, we report failure, if there is a pass, we report that and stop retrying. The failed-then-passed attempts do not affect the final result of the run, but can be inspected in the web interface. See [our docs](https://help.rainforestqa.com/docs/test-retries) for more detail.
```
magni-, line 363: Is nothing exposed via the JUnit report? I thought I saw a PR adding that

maciejgryka: Yes, there's https://github.com/rainforestapp/Rainforest/pull/19451/files - do you mean we should just mention it here?

PR #448, opened 2022-02-11
[RF-22876] Support downloading and uploading existing wisp tests

pyromaniackeca:
Adds support for download and uploading RFA tests, by storing their WISP in json format along with the test id and (arbitrarily) title. Motivation for doing this is allowing users to backup and restore their RFA tests via CLI.

example_rfa_test.json
``` json
{
  "id": 1234,
  "title": "some_title",
  "wisp": {. . .}
}
```

Stored files use the same naming convention as RFML aside from their extension being `.json`. This was chosen arbitrarily and maybe should be changed.

New wisp support was added to the existing upload and download commands as to reuse existing code for handling files and the API.

This PR should not break any existing behavior and only adds support for RFA tests, before they would blow up.
Tested by backing up and restoring a run group of 3 tests (rfml, rfa, rfa with embedded test) by run group id and individual ids.

Reviewing commit by commit might be easier, each commit should be self contained.


rf_tests.go
```diff
@@ -342,13 +342,20 @@ func deleteRFML(c cliContext) error {
 	return nil
 }
 
-// uploadRFML is a wrapper around test creating/updating functions
-func uploadRFML(c cliContext, api rfmlAPI) error {
+// uploadTests is a wrapper around test creating/updating functions
+func uploadRFTests(c cliContext, api rfAPI) error {
```
jbarber, line 346: Function name differs from comment?
```suggestion
func uploadTests(c cliContext, api rfAPI) error {
```

rf_tests.go
```diff
@@ -342,13 +342,20 @@ func deleteRFML(c cliContext) error {
...
 	if c.Bool("synchronous-upload") {
 		rfmlUploadConcurrency = 1
 	}
 	if path := c.Args().First(); path != "" {
-		err := uploadSingleRFMLFile(path)
+		var err error
+
+		if strings.HasSuffix(path, ".rfml") {
```
jbarber, line 353: It'd be nice to determine the file type by it's contents rather than file extension? Could we try reading them as WISP and if that fails fall back to RFML (or vice versa)?

rf_tests.go
```diff
@@ -342,13 +342,20 @@ func deleteRFML(c cliContext) error {
...
+			err = uploadSingleRFMLFile(path)
+		} else {
+			err = uploadSingleWispFile(path)
```
jbarber, line 356: I'm aware this isn't your naming convention, and we might not want to make this change as part of this PR - but including the word `Single` in the function name is completely redundant?

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
 	if err != nil {
 		return cli.NewExitError(err.Error(), 1)
 	}
+
+	wisp_tests, err := readWispFiles([]string{c.String("test-folder")})
+	if err != nil {
+		return cli.NewExitError(err.Error(), 1)
+	}
+	err = uploadWispFiles(wisp_tests, false, api)
```
jbarber, line 377: This function doesn't deal with files at all, so is misnamed?
```suggestion
	err = uploadWisps(wisp_tests, false, api)
```

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
...
+		wisp, err := readWispFile(filePath)
+		if err != nil {
+			return nil, err
+		}
+		tests = append(tests, wisp)
+	}
+	return tests, nil
+}
+
+func uploadWispFiles(tests []*rainforest.WispJson, localOnly bool, api rfAPI) error {
```
jbarber, line 464: `localOnly` isn't used?

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
...
+				continue
+			} else {
+				log.Printf("%s is not a valid json file", file)
+				continue
+			}
+		}
+
+		// We have a directory, walk through and find wisp files
+		err = filepath.Walk(file, func(path string, f os.FileInfo, err error) error {
+			if strings.HasSuffix(path, ".json") {
```
jbarber, line 437: Bad stuff will happen if there's a sub-directory with a name of `foo.json`?

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
...
+		}
+		seenPaths[filePath] = true
+		wisp, err := readWispFile(filePath)
+		if err != nil {
+			return nil, err
+		}
+		tests = append(tests, wisp)
+	}
+	return tests, nil
+}
```
jbarber, line 462: I think this function is doing too much as it's finding all the Wisp looking tests and it's also parsing them?

Perhaps split it into two function, the first does the file discovery and returns a list of (unique) paths, the second maps over this list of paths and returns an array of `WispJson` objects.

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
...
+	wispReader := rainforest.NewWispReader(file)
+	wisp, err := wispReader.Read()
+	if err != nil {
+		return nil, fileParseError{filePath, err}
+	}
+
+	return wisp, err
+}
+
+// readWispFiles takes in a list of files and/or directories and a list of tags
```
jbarber, line 415: What are tags - copypasta from some other comment?

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
...
+				fileList = append(fileList, file)
+				continue
+			} else {
+				log.Printf("%s is not a valid json file", file)
+				continue
+			}
+		}
+
+		// We have a directory, walk through and find wisp files
+		err = filepath.Walk(file, func(path string, f os.FileInfo, err error) error {
```
jbarber, line 436: https://pkg.go.dev/path/filepath#Walk says `WalkDir` is more efficient than `Walk` - I don't know if we care.

rainforest/tests.go
```diff
@@ -128,6 +130,56 @@ type testElementDetails struct {
 	Elements []testElement `json:"elements,omitempty"` // for embedded tests
 }
 
+type Wisp struct {
+	Version string `json:"version"`
+	Verbs   []Verb `json:"verbs"`
+}
```
jbarber, line 136: How do we handle different versions of Wisp in the future?

rf_tests.go
```diff
@@ -362,6 +369,108 @@ func uploadRFML(c cliContext, api rfmlAPI) error {
...
+		return cli.NewExitError(err.Error(), 1)
+	}
+
+	return nil
+}
+
+// uploadSingleWispFile uploads wisp file syntax by
+// trying to parse the file and sending any parse errors to the caller
+func uploadSingleWispFile(filePath string) error {
+	wisp, _ := readWispFile(filePath)
```
jbarber, line 388: Why ignore errors here? Possibly it's safe because if there was an error than `wisp` will have be an empty struct with `TestID = 0`, and this will get caught by `UpdateWisp` - but we should fail quicker?

rainforest/tests.go
```diff
@@ -128,6 +130,56 @@ type testElementDetails struct {
...
+	Confirmation   *string `json:"confirmation,omitempty"`
+}
+
+type Noun struct {
+	Type          string  `json:"type"`
+	ID            *int    `json:"id,omitempty"`
+	Element       *string `json:"element,omitempty"`
+	MatchMode     *string `json:"match_mode,omitempty"`
+	FocusX        *int    `json:"focus_x,omitempty"`
+	FocusY        *int    `   json:"focus_y,omitempty"`
```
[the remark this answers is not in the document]
pyromaniackeca: Missed it, thanks!

rf_tests.go
pyromaniackeca: Yes, I'll remove it. It doesn't work with wisp at least atm, not sure if that's something we'll add to match RFML.

rf_tests.go
pyromaniackeca: Yes, that is why I decided to ignore the error here. I wanted to have one less `if err != nil { . . . }` in the function, but failing quicker makes sense, plus you'll also get a more descriptive error to the actual issue.

rf_tests.go
pyromaniackeca: Do you prefer `uploadtests`? I initially went with `uploadTests` and changed it to `uploadRFTests`. I think it doesn't make a huge difference here, but I think naming file `rf_tests` rather than the generic `tests` is probably better and wanted to make the function names match.

rf_tests.go
pyromaniackeca: Was mirroring the naming of the RFML functions, but makes perfect sense. 👍

rainforest-cli.go
```diff
@@ -400,10 +400,10 @@ func main() {
 			Name: "download",
 			// Left for legacy reason, should nuke?
 			Aliases:      []string{"export"},
-			Usage:        "Download your remote Rainforest tests to RFML",
+			Usage:        "Download your remote Rainforest tests",
 			OnUsageError: onCommandUsageErrorHandler("download"),
 			ArgsUsage:    "[test IDs]",
-			Description: "Download your remote tests from Rainforest to RFML. " +
+			Description: "Download your remote tests from Rainforest to RFML or Wisp. " +
```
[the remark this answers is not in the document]
pyromaniackeca: Will remove remote for both, as for the latter, I'm not sure which is better. Maybe the value now is making users aware that not all files will be stored the same?

rf_tests.go
jbarber: Yeah, I wondered about this as well and wasn't sure. But thinking about it, this function is in the main package and is not exported, so there shouldn't be any clashes with any functions we might want to have in the future, and it seems unlikely we'll want to upload tests other than RF tests, so make it `uploadTests`? But it's nbd - because it's not like we can't change it in the future anyway.

rainforest-cli.go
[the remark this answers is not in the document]
jbarber: Yeah, but they can't do anything about it, so it doesn't really matter to them?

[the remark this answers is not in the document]
pyromaniackeca: Makes sense, I don't have strong feelings about keeping it so will remove as well.

rf_tests.go
pyromaniackeca: I'll see how to check if something is a dir as well.

rf_tests.go
pyromaniackeca: I'll see if we depend on the functionality removed from `Walk`, other than that it seems easy to change, so why not?

rainforest/tests.go
jbarber: Have you thought about how and when we update Wisp to have different verbs (and so the Wisp schema changes), how will the CLI support having different versions (either stored locally, or downloaded from the product)?

It is necessary to support prior versions so we can update the CLI+Wisp without everything breaking whenever we update.

pyromaniackeca: Currently the only thing that would need updating is the wisp structures.
If we add new fields it would just ignore them, so the failure would come from trying to reupload a downloaded test.
To fix this we would need to add any new field (or any old removed field from prior versions if any)as an optional field in the wisp structs.

Alternatively if we don't want to do that, we could accept the entire wisp as a json blob and store/reupload that.
This would require us then to delegate any and all validation to the backend as writing and maintaining a validator defeats the purpose of removing the structs.

I guess it depends on how involved we want the CLI to be in the contents of the wisp. Alternatively we could also expose an endpoint(somewhere) to validate wisp, which could allow us to accept a json blob on the client side and still be able to validate it when necessary.

While writing this, I realized that I was working on the assumption that client side validation will be required at some point(currently we're not really doing it, but the structures are a first step), so if that assumption is wrong, the json blob might be the right choice.

rf_tests.go
pyromaniackeca: Seems we don't need to use `Walk` over `WalkDir`, switched.

rf_tests.go
pyromaniackeca: @jbarber I've decided to leave this as is. `rfml` already expects the extension to be correct in a few places, but my main issue is that just a simple fallback can potentially hide an error with the file.

Eg:
- Try to upload invalid wisp
- Get error wisp is invalid
- Fallback to try to upload rfml
- Get error rfml is invalid
- return rfml invalid error to user for an invalid wisp

The argument there would be to add better error handling, but that's a bit more involved than I would like for this PR.
If we want that, I would like to do it in a separate PR and revisit how rfml is handled as well.

The other thing you mentioned is just changing the extensions to `.wisp`, that's a simple change to make now.

rf_tests.go
pyromaniackeca: Handling this is a bit involved since the way the code is structured now we would need to handle it in a few places.
Atm this blows up for both `rfml` and `wisp`. Wisp says that the json while is a dir, while rfml lists why the rfml is invalid.

I think wisp blows up relatively sensibly for now and if we want to handle this properly, we should probably restructure the code in a separate PR.

rainforest/tests.go
pyromaniackeca: @jbarber I tried changing the structures to a json blob as per our discussion, but it doesn't just work. The blocker here is that actions have an id which blows up backend when trying to update. Backend checks if an id is present and if it is it tries to find an element matching the id. Because some elements might be deleted, it can't find them and raises an error.

This worked with structures as I was ignoring the id, so it would never be send to backend and backend would create a new element. We discussed before that we might want to change backend to not blow up if there is a provided id, but the element is deleted.

My suggestion would be to either ship as is for now and revisit later, or update backend to not blow up when an id is provided for an element, but the element is deleted.
Updating backend would also have the added benefit of reusing existing elements when possible instead of always creating a new one.

After a quick check of the code, it seems like it wouldn't take much work to make this change backend.

rf_tests.go
pyromaniackeca: I ended up separating intro three functions. One finds the files, second removes dups and third one parses the json.
I plan to further refactor this in a later PR to reuse the dup filter and probably the file discovery as well for rfml and wisp to reduce code duplication.

rainforest/tests.go
```diff
@@ -128,6 +130,56 @@ type testElementDetails struct {
 	Elements []testElement `json:"elements,omitempty"` // for embedded tests
 }
 
+type Wisp struct {
```
magni-, line 133: This won't solve the issue of outdated clients using an outdated schema, but why not reuse the JSON schema here rather than manually define the `Wisp` type?

rainforest/wisp.go
```diff
@@ -0,0 +1,75 @@
...
+	"io"
+	"log"
+)
+
+// WispWriter writes test's id, title and wisp to a given file in json format.
+type WispWriter struct {
+	w *bufio.Writer
+}
+
+// WispReader reads from a test's id, title and wisp from a formatted json file.
```
magni-, line 15:
```
// WispReader reads a test's id, title and wisp from a formatted json file.
```

magni- (commented):
> we could accept the entire wisp as a json blob and store/reupload that.
This would require us then to delegate any and all validation to the backend as writing and maintaining a validator defeats the purpose of removing the structs.
>
> I guess it depends on how involved we want the CLI to be in the contents of the wisp. Alternatively we could also expose an endpoint(somewhere) to validate wisp, which could allow us to accept a json blob on the client side and still be able to validate it when necessary.
>
> While writing this, I realized that I was working on the assumption that client side validation will be required at some point(currently we're not really doing it, but the structures are a first step), so if that assumption is wrong, the json blob might be the right choice.

I think this is the right choice (to start at least). That way a Wisp update won't require a CLI update.

rainforest/tests_test.go
```diff
@@ -298,6 +298,39 @@ func TestGetTest(t *testing.T) {
...
+			Title:  "A test",
+		}
+		options := r.URL.Query().Get("options[]")
+		if options != "wisp" {
+			t.Error("Options param wasn't set to wisp.")
+		}
+
+		exclude := r.URL.Query().Get("exclude[]")
+		if exclude != "elements" {
+			t.Error("Options param wasn't set to elements.")
```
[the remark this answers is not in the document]
pyromaniackeca: Thanks! 👍

rainforest/tests.go
pyromaniackeca: I tried a few go solutions, but either couldn't make them work or they weren't working properly. Ultimately it seemed like more of a pain(at least for this initial version) than writing it by hand.

rainforest/wisp.go
pyromaniackeca: Thanks! 👍

rainforest/tests.go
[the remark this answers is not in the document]
jbarber: Weird that `go fmt` doesn't have an opinion on it.

rf_tests.go
jbarber: I think we covered this in our 1:1 - but to recap, the backend is the source of truth, so I think it's ok to delegate validation of WISP to it and that means the workflow would be:
* Try to parse as JSON, on failure parse as RFML (or vice versa)
* Do stuff with all the parsed files

IMO it's fine to ignore in this PR.

jbarber (approved):
Seems good. @maciejgryka is the CLI interface ok for us to commit to supporting? (I think it is, but want to check before we ship this because rolling it back would be a breaking change for any users who might start depending on it)

rf_tests.go
pyromaniackeca: I'll ignore it for now. My main issue is that I believe we will be providing a misleading error in some circumstances(like the one I described above).

I will address this in a cleanup PR with a proposed solution and we can discuss further there.

rainforest/tests.go
```diff
@@ -525,3 +587,27 @@ func (c *Client) UpdateTest(test *RFTest) error {
 	}
 	return nil
 }
+
+// UpdateWisp updates existing test's wisp and title on RF, requires WispJson struct to be prepared to upload using helpers
+func (c *Client) UpdateWisp(wisp *WispJson) error {
+	if wisp.TestID == 0 {
+		return errors.New("Couldn't update the test TestID not specified in RFTest")
+	}
```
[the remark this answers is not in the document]
pyromaniackeca: Backend does blow up, but I would prefer to give a better error message than `Can't find test with id 0`.

My reasoning for putting it here was to preemt the backend error and translate to a more informative message, but I agree it would be better placed elsewhere. I'll move it.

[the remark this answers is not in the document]
pyromaniackeca: Actually this was copy-pasta from from the `UpdateTest` function and I assume the intention of the original author was to provide a better error message. I'll leave this for the cleanup PR as well.

maciejgryka:
Just had a chat with @pyromaniackeca, summarizing here.

I don't think we should commit to supporting this @jbarber - it's likely we'll have a different solution to the branching problem in the future and supporting two ways of doing the same thing seems like unnecessary pain. As things are, I don't foresee us pursuing this way of supporting branching on the product side.

Now, this PR might still be useful internally as a way to solve some of our pains temporarily. We discussed two ways of doing this:

1. Keeping the functionality as part of the `download`/`upload` commands, but only work with VE (Visual Editor) tests when the `--experimental` flag is provided.
2. Adding new undocumented/not-officially-supported commands for dealing with VE tests.

I'd prefer option 2 because it feels cleaner/more obviously  unsupported, but I'm OK with 1. too if it's easier to ship.

P.S. There's a fix included here to make the `download` command not blow up when used with mixed-crowd run groups. We should ship that fix to clients, independently of the ability to download VE tests.

pyromaniackeca:
Decided not to pursue this further. Closing this PR and will open a separate PR to keep some of the cleanup as well as an improvement to the error given when wisp stored tests are being downloaded.

PR #450, opened 2022-02-23
[RF-23592] Better handle trying to download unsupported tests

pyromaniackeca:
- Provide better error when user tries to download an unsupported tests
- Rename rfml files to better match naming between namespaces
- Add `.tool-versions` file.


tests.go
```diff
@@ -342,18 +341,24 @@ func deleteRFML(c cliContext) error {
...
+func uploadTests(c cliContext, api rfAPI) error {
 	if c.Bool("synchronous-upload") {
 		rfmlUploadConcurrency = 1
 	}
+
 	if path := c.Args().First(); path != "" {
-		err := uploadSingleRFMLFile(path)
+		var err error
+
+    err = uploadSingleRFMLFile(path)
```
jbarber, line 353: Weird formatting, or is it GH's diff doing something weird?

jbarber: Why not
```suggestion
err := uploadSingleRFMLFile(path)
```
?

tests.go
```diff
@@ -634,44 +659,45 @@ func downloadRFML(c cliContext, client rfmlAPI) error {
...
+	for i := 0; i < len(testInfos); i++ {
 		select {
 		case err = <-errorsChan:
 			return cli.NewExitError(err.Error(), 1)
 		case test := <-testChan:
-			err = test.PrepareToWriteAsRFML(*testIDCollection, c.Bool("flatten-steps"))
-			if err != nil {
-				return cli.NewExitError(err.Error(), 1)
-			}
+      err = test.PrepareToWriteAsRFML(*testIDCollection, c.Bool("flatten-steps"))
```
jbarber, line 667: Indent is wrong?

tests.go
```diff
@@ -634,44 +659,45 @@ func downloadRFML(c cliContext, client rfmlAPI) error {
...
+			file, err = os.Create(filePath)
 			if err != nil {
 				return cli.NewExitError(err.Error(), 1)
 			}
 
-			writer := rainforest.NewRFMLWriter(file)
-			err = writer.WriteRFMLTest(test)
+      writer := rainforest.NewRFMLWriter(file)
+      err = writer.WriteRFMLTest(test)
+
```
jbarber, line 685: Indentation

jbarber (approved):
Few comments but seems good.

After you ship this can you have a look into the CI/CD config please? I'm pretty sure we have a formatting check that I'd expect to catch the weird indentation, so it'd be good to have that working.

tests.go
pyromaniackeca: Thanks that's a leftover from when there were to options and I needed to declare the `var`, before setting it in an `if/else`.

Will change.

pyromaniackeca:
@jbarber I've pushed a commit to update the changelog as well. In regards to checking formatting, we don't have it in CI. I'll open a separate PR to add it, as to keep this one more focused. There might be some discussion points on how exactly we want to add it.


One thing I'm not 100% about is should I include the `.tool-versions` commit in the changelog?

jbarber:
> @jbarber I've pushed a commit to update the changelog as well. In regards to checking formatting, we don't have it in CI. I'll open a separate PR to add it, as to keep this one more focused. There might be some discussion points on how exactly we want to add it.

Yep, sounds good.

> One thing I'm not 100% about is should I include the `.tool-versions` commit in the changelog?

I don't think so, IMO the changelog is targeted at users of the CLI and they won't care about us changing things like that.

pyromaniackeca:
> > One thing I'm not 100% about is should I include the `.tool-versions` commit in the changelog?
> 
> I don't think so, IMO the changelog is targeted at users of the CLI and they won't care about us changing things like that.

I agree, thanks for the second opinion.

PR #451, opened 2022-02-24
[RF-23601] Enforce gofmt code formatting

pyromaniackeca:
Add circleci check for correct code formatting.

Alternatively we could also use: [golangci-lint](https://github.com/golangci/golangci-lint)
The reason I opted not to is because it does a lot more, so it's more involved to setup correctly for us.

We would probably want to add this in eventually though, it also has a github actions integration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,13 @@ jobs:
       - run:
           name: Check GoReleaser Config
           command: goreleaser check
+  validate_formatting:
+    executor: linux
+    steps:
+      - checkout
+      - run:
+          name: Validate code formatting
+          command: ./script/validate_formatting
   test:
     executor: linux
     steps:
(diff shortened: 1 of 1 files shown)


script/validate_formatting
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Find incorrectly formatted files
+unformatted=$(gofmt -l .)
```
pyromaniackeca, line 4: We only had a few simple offences, so I opted to fix them and run for everything to keep things simpler. We could also change this to only run on differences.

rainforest-cli.go
```diff
@@ -152,7 +152,7 @@ func main() {
 			Aliases:      []string{"r"},
 			Usage:        "Run your tests on Rainforest",
 			OnUsageError: onCommandUsageErrorHandler("run"),
-			Action:       func(c *cli.Context) error {
+			Action: func(c *cli.Context) error {
```
CChanHY, line 155: Is it right that some are "line up" formatted and some are not? `func` is an exception? That would be understandable though.

pyromaniackeca: It was autofixed and the check isn't complaining, so that's probably the convention.

jaytennier (approved):
My bash-fu is pretty weak but it all looks good to me

PR #458, opened 2022-09-20
[RF-24889] Add `--execution-method` flag, deprecate `--crowd`

magni-:
Regardless of which flag users use, we'll send `execution_method` to the backend.

This PR also updates the output of the `run_groups` command to show the execution method rather than the crowd.

### DIFF
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -13,7 +13,7 @@ import (
 
 const (
 	// Version of the app in SemVer
-	version = "3.0.0"
+	version = "3.1.0"
 	// This is the default spec folder for RFML tests
 	defaultSpecFolder = "./spec/rainforest"
 )
--- a/rainforest-cli.go
+++ b/rainforest-cli.go
@@ -206,9 +206,13 @@ func main() {
 					Usage: "Run your tests using specified `ENVIRONMENT`. Otherwise it will use your default one.",
 				},
 				cli.StringFlag{
-					Name: "crowd",
-					Usage: "Run your tests using specified `CROWD`. Available choices are: default, automation, automation_and_crowd " +
-						"or on_premise_crowd. Contact your CSM for more details.",
+					Name: "execution-method",
+					Usage: "Run your tests using specified `EXECUTION_METHOD`. Available choices are: crowd, automation, automation_and_crowd, " +
+						"and on_premise. Some methods may not be available for your account. For more information, contact us at help@rainforestqa.com.",
+				},
+				cli.StringFlag{
+					Name:  "crowd",
+					Usage: "DEPRECATED: Use --execution-method instead",
 				},
 				cli.StringFlag{
 					Name: "conflict",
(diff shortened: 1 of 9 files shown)


rainforest-cli.go
```diff
@@ -206,9 +206,13 @@ func main() {
 					Usage: "Run your tests using specified `ENVIRONMENT`. Otherwise it will use your default one.",
 				},
 				cli.StringFlag{
-					Name: "crowd",
-					Usage: "Run your tests using specified `CROWD`. Available choices are: default, automation, automation_and_crowd " +
-						"or on_premise_crowd. Contact your CSM for more details.",
+					Name: "execution-method",
+					Usage: "Run your tests using specified `EXECUTION_METHOD`. Available choices are: crowd, automation, automation_and_crowd " +
+						"or on_premise. Some methods may not be available for your account.",
```
CChanHY, line 211:
```suggestion
						"and on_premise. Some methods may not be available for your account.",
```
I think this sounds more natural because `on_premise` is _also_ part of the choices.

Should we add this here too as in the README:  "For more information, contact us at help@rainforestqa.com"

rainforest/runner.go
```diff
@@ -94,8 +94,8 @@ func validateRerunParams(params RunParams) error {
 	if params.SiteID != 0 {
 		return errors.New("Site cannot be specified for rerun")
 	}
-	if params.Crowd != "" {
-		return errors.New("Crowd cannot be specified for rerun")
+	if params.ExecutionMethod != "" {
+		return errors.New("Execution method cannot be specified for rerun")
```
CChanHY, line 98:
```suggestion
		return errors.New("Execution Method cannot be specified for rerun")
```
Because "Execution Method" together is a "thing", like "Run Group" below.

runner.go
```diff
@@ -530,6 +530,32 @@ func getPlatforms(c cliContext) []string {
 	return expandStringSlice(platforms)
 }
 
+func getExecutionMethod(c cliContext) (string, error) {
+	var crowd string
+	if crowd = c.String("crowd"); crowd != "" {
+		fmt.Println("RF CLI Deprecation: --crowd is deprecated, use --execution-method instead")
+	}
```
CChanHY, line 537: Maybe put this in a `getCrowd` method that always gives this error, so the deprecation code is not infecting this method?

runner.go
```diff
@@ -530,6 +530,32 @@ func getPlatforms(c cliContext) []string {
...
+	executionMethod := c.String("execution-method")
+
+	if crowd != "" && executionMethod != "" {
+		return "", errors.New("execution-method and crowd are mutually exclusive")
```
CChanHY, line 541: Wait, what? I thought 'crowd' is just deprecated?

runner.go
magni-: I thought about that, but the whole method will need to be rewritten anyways when `crowd` is removed.

runner.go
magni-: `--execution-method` replaces `crowd`. The two parameters are already mutually exclusive at the API level, this just bubbles that existing validation up to the CLI.

CChanHY (approved):
👍

PR #459, opened 2022-10-05
RF-24826 Add snippet support to RFML files

jaytennier:
Edit: I just marked this as DNM because we don't want this released until we coordinate with Domino on a time that works for them. As of this moment @pyromaniackeca has been asked to shepherd this release out when it makes sense to do so.

Adds support for downloading and uploading tests and snippets via RFML. This introduces a new RFML field `type` that can be either `test` for top-level tests or `snippet` for the new reusable snippet-type tests (formerly "embedded tests").

* Running `rainforest download` will make sure all existing tests have their `type` set correctly.
* Running `rainforest new` will generate a new RFML file with its `type` set to `test`. You can manually change that to `snippet` if you need to embed it in another test/snippet.
* Running `rainforest upload` will update the tests as usual and create any tests / snippets depending on their type. Note: changing the `type` in the RFML file is not supported. The rest of test will still get updated correctly but its type will not change.

### DIFF
--- a/rainforest/rfml_test.go
+++ b/rainforest/rfml_test.go
@@ -182,6 +182,29 @@ func TestReadAll(t *testing.T) {
 		t.Errorf("Incorrect test priority. Got %v, Want empty string", rfTest.Priority)
 	}
 
+	// Test type is present
+	testText = fmt.Sprintf(`#! %v
+# title: %v
+# start_uri: %v
+# type: %v
+`,
+		validTestValues.RFMLID,
+		validTestValues.Title,
+		validTestValues.StartURI,
+		"test",
+	)
+
+	r = strings.NewReader(testText)
+	reader = NewRFMLReader(r)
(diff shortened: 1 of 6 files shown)


rainforest/rfml_test.go
```diff
@@ -182,6 +182,29 @@ func TestReadAll(t *testing.T) {
 		t.Errorf("Incorrect test priority. Got %v, Want empty string", rfTest.Priority)
 	}
 
+	// Test type is present
+	testText = fmt.Sprintf(`#! %v
+# title: %v
+# start_uri: %v
+# type: %v
+`,
```
pyromaniackeca, line 190: Formatting seems broken here, but if the check passes, it's probably correct. 🤷

pyromaniackeca (approved):
Echoing @magni- suggestions. I think it makes sense to add a `--snippet` option when creating and blowing up if trying to change the test type.

PR #460, opened 2022-10-11
[DNM] [RF-25462] Add command to create new branch

pyromaniackeca:
Adds a command to create a new branch.

@maciejgryka Can you check the copy of the output/help command?
Also do we want to hide these commands for now as they're not available to everyone?

@magni- Do you have any tips for improving the code or the tests?

I'll also not be merging this until #459 has been released to make things simpler. That should be happening this week.

### DIFF
--- a/branches.go
+++ b/branches.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/rainforestapp/rainforest-cli/rainforest"
+	"github.com/urfave/cli"
+)
+
+type branchAPI interface {
+	CreateBranch(*rainforest.Branch) error
+}
+
+func newBranch(c cliContext, api branchAPI) error {
+	name := c.Args().First()
+	name = strings.TrimSpace(name)
+
+	if name == "" {
+		return cli.NewExitError("Branch name cannot be blank", 1)
+	}
+
+	branch := rainforest.Branch{
+		Name: name,
+	}
+
+	err := api.CreateBranch(&branch)
+
+	if err != nil {
+		return cli.NewExitError(err.Error(), 1)
+	}
+
+	fmt.Printf("Created branch %q.\n", name)
+	return nil
+}
--- a/rainforest/branches.go
+++ b/rainforest/branches.go
@@ -1,0 +1,25 @@
+package rainforest
+
+// Branch is a struct representing a Rainforest Branch
+type Branch struct {
+	ID   int    `json:"id,omitempty"`
+	Name string `json:"name"`
+}
+
+// CreateBranch creates new branch on RF, requires Branch struct to be prepared to upload using helpers
+func (c *Client) CreateBranch(branch *Branch) error {
+	// Prepare request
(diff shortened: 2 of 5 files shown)


branches.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/urfave/cli"
+)
+
+type branchAPI interface {
+	CreateBranch(*rainforest.Branch) error
+}
+
+func newBranch(c cliContext, api branchAPI) error {
+	name := c.Args().First()
+	name = strings.TrimSpace(name)
```
pyromaniackeca, line 16: We might not need this line as it seems like the cli context already does it(which makes sense), but I wasn't able to find a confirmation of this. I didn't bother going trough the source for the sake of time. @magni- Any thoughts?

magni-: We don't need this for (`$` to show EOL)
```
./rainforest branches new      some-branch-name    $
```
But if we want to disallow precending/trailing whitespace, we'll need it to trim cases such as
```
./rainforest branches new "  name with spaces before and after   "$
```
(but maybe that trimming should happen on the backend instead?)

pyromaniackeca: Good point on handling this in the backend. Will remove it here and add to the backend if we're not already doing it. 👍

pyromaniackeca: On second thought I'll leave this in to make the output nicer. `"branchName"` vs `" branchName "`.
Atm we don't do trimming on the backend. I'll add it there as well.

rainforest/branches.go
```diff
@@ -0,0 +1,25 @@
+package rainforest
+
+// Branch is a struct representing a Rainforest Branch
+type Branch struct {
+	ID   int    `json:"id"`
+	Name string `json:"name"`
+}
+
+// CreateBranch creates new branch on RF, requires Branch struct to be prepared to upload using helpers
+func (c *Client) CreateBranch(branch *Branch) error {
```
magni-, line 10: Should we add tests for this? See the tests in `rainforest/runner_test.go` for example

magni- (approved):

rainforest/branches.go
pyromaniackeca: I decided not to initially as there isn't really much to test here and I used the `rainforest/tests_test.go` as a reference where this isn't tested explicitly.

After taking a look at the example you suggested, I added a simple test to check the body of the request.
Thanks for the suggestion.

magni-:
Should we also add documentation for this to the `README`? Maybe link to https://help.rainforestqa.com/docs/using-branching ? It doesn't matter that the feature is not available to everyone right now IMO, we also mention things like on-prem in there.

pyromaniackeca:
> Should we also add documentation for this to the `README`? Maybe link to https://help.rainforestqa.com/docs/using-branching ? It doesn't matter that the feature is not available to everyone right now IMO, we also mention things like on-prem in there.

Will do, we plan to return an error from the backend saying that users can request early activation. So it makes sense to document it properly.

pyromaniackeca:
Closing in favor of #462 in order to simplify shipping this.

PR #462, opened 2022-10-20
[RF-25461] Add branching support to rainforest CLI

pyromaniackeca:
This PR adds the following commands to the cli:
- `branch new <name>` creates a new branch
- `branch delete <name>` deletes an existing branch
- `branch merge <name>` merges an existing branch

It also adds the `--branch <name>` option to the `run` and `upload` commands enabling them to be applied to a branch.

I've added one command per commit and a separate refactoring commit, so reviewing per commit might be easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,12 @@ Run individual tests in the foreground and report.
 rainforest run <test_id1> <test_id2>
 ```
 
+Run all tests on a branch.
+
+```bash
+rainforest run all --branch branch-name
+```
+
 Run a run group.
 
 ⚠️ This uses the configuration defined in the run group (environment, platforms, execution method, location). If you wish to run tests from a run group without using the run group's configuration, you will need to use the Rainforest API directly, passing a `run_group_id` parameter to [the `POST /runs` endpoint](https://help.rainforestqa.com/reference/post-runs). ⚠️
@@ -156,6 +162,12 @@ Upload a specific test to Rainforest
 
 ```bash
 rainforest upload /path/to/test/file.rfml
+```
+
+Upload a specific test to Rainforest on a branch
+
+```bash
+rainforest upload --branch branch-name /path/to/test/file.rfml
 ```
 
 Remove RFML file and remove test from Rainforest test suite.
@@ -249,6 +261,23 @@ rainforest csv-upload --import-variable-name my_variable PATH/TO/CSV.csv
 Upload a CSV to update an existing tabular variables.
 ```bash
 rainforest csv-upload --import-variable-name my_variable --overwrite-variable PATH/TO/CSV.csv
+```
+
+#### Managing branches
+
+Create a new branch.
+```bash
+rainforest branch new branch-name
+```
+
+Delete an existing branch.
+```bash
+rainforest branch delete branch-name
+```
+
+Merge an existing branch into the `main` branch.
+```bash
+rainforest branch merge branch-name
 ```
 
 #### Uploading Mobile Apps
--- a/branches.go
+++ b/branches.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/rainforestapp/rainforest-cli/rainforest"
+	"github.com/urfave/cli"
+)
+
+type branchAPI interface {
+	CreateBranch(*rainforest.Branch) error
+}
+
+func newBranch(c cliContext, api branchAPI) error {
+	name := c.Args().First()
+	name = strings.TrimSpace(name)
+
+	if name == "" {
+		return cli.NewExitError("Branch name cannot be blank", 1)
+	}
+
+	branch := rainforest.Branch{
+		Name: name,
+	}
+
+	err := api.CreateBranch(&branch)
+
+	if err != nil {
+		return cli.NewExitError(err.Error(), 1)
+	}
+
+	fmt.Printf("Created branch %q.\n", name)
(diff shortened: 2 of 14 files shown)


rainforest/branches.go
```diff
@@ -1,11 +1,65 @@
...
+
+		var branchResp []Branch
+		_, err = c.Do(req, &branchResp)
+		if err != nil {
+			return nil, err
+		}
+
+		branches = append(branches, branchResp...)
+
+		totalPagesHeader := c.LastResponseHeaders.Get("X-Total-Pages")
```
magni-, line 45: Can we DRY up this auto-paging functionality?

branches.go
```diff
@@ -36,6 +37,36 @@ func newBranch(c cliContext, api branchAPI) error {
...
+	name := c.Args().First()
+	name = strings.TrimSpace(name)
+
+	if name == "" {
+		return cli.NewExitError("Branch name cannot be blank", 1)
+	}
+
+	branches, err := api.GetBranches(name)
+
+	if err != nil {
```
magni-, line 51: Same question about preventing attempting to merge `main` locally

rainforest/tests.go
```diff
@@ -508,17 +508,19 @@ func (c *Client) CreateTest(test *RFTest) error {
...
 // UpdateTest updates existing test on RF, requires RFTest struct to be prepared to upload using helpers
-func (c *Client) UpdateTest(test *RFTest) error {
+func (c *Client) UpdateTest(test *RFTest, branchID int) error {
 	if test.TestID == 0 {
 		return errors.New("Couldn't update the test TestID not specified in RFTest")
 	}
 
+	url := fmt.Sprintf("tests/%d?slim=true", test.TestID)
+
+	if branchID != 0 {
```
magni-, line 518: Could we use an explicit constant rather than a magic `0` here? Something like
```go
if branchID != NO_BRANCH {
```

rainforest/tests_test.go
```diff
@@ -520,7 +547,7 @@ func TestUpdateTest(t *testing.T) {
 		}
 	})
 
-	err = client.UpdateTest(&rfTest)
+	err = client.UpdateTest(&rfTest, 0)
```
magni-, line 550: Could we add a comment documenting why this is hardcoded for now?

tests_test.go
```diff
@@ -393,6 +394,23 @@ func TestUploadTests(t *testing.T) {
 	if err != nil {
 		t.Fatal(err.Error())
 	}
+
+	// with a branch
+	testAPI.handleUpdateTest = func(rfTest *rainforest.RFTest, branchID int) {
+		if branchID != 1 {
+			t.Errorf("Incorrect value for branchID. Expected 123, Got %v", branchID)
```
magni-, line 401: 👀

runner.go
```diff
@@ -65,16 +65,30 @@ func (r *runner) startRun(c cliContext) error {
...
 
-	var localTests []*rainforest.RFTest
 	var err error
+	var branchID int
+	branchName := c.String("branch")
+	branchName = strings.TrimSpace(branchName)
+	if branchName != "" {
+		branchID, err = getBranchID(branchName, r.client)
+	} else {
+		branchID = 0
```
magni-, line 75: Same with regards to magic number vs constant

README.md
```diff
@@ -251,6 +263,23 @@ Upload a CSV to update an existing tabular variables.
...
+```bash
+rainforest branch new branch-name
+```
+
+Delete an existing branch.
+```bash
+rainforest branch delete branch-name
+```
+
+Merge an existing branch into main.
```
magni-, line 278: Suggestion: ... into the `main` branch

magni- (approved):

tests_test.go
pyromaniackeca: 🙈

rainforest/tests_test.go
pyromaniackeca: Updated it to use the constant.

branches.go
```diff
@@ -33,3 +35,33 @@ func newBranch(c cliContext, api branchAPI) error {
...
+		return cli.NewExitError(err.Error(), 1)
+	}
+
+	if len(branches) == 0 {
+		return cli.NewExitError("Cannot find branch", 1)
+	}
+
+	branch := branches[0]
+
+	err = api.DeleteBranch(branch.ID)
```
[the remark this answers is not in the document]
pyromaniackeca: I'm not sure, the benefit of relying on the backend is that if the rules change we don't need to change them here as well. The rules are very unlikely to change for this though.

The benefit of doing it here is it saves us an API request and makes it easier to customize the error? Anything else?

I've created a [ticket](https://rainforest.atlassian.net/browse/RF-25651) and added it as a point for consideration. I won't do it for now as the difference seems minor and not worth delaying this.

rainforest/branches.go
pyromaniackeca: Yep but I won't do it as part of this PR. Added a [ticket](https://rainforest.atlassian.net/browse/RF-25651).

branches.go
pyromaniackeca: Same points as for handling deleting main. Added to the [ticket](https://rainforest.atlassian.net/browse/RF-25651).

pyromaniackeca:
@magni- I've addressed the comments and did some minor refactors to improve the code and tests. Thanks for the review!

PR #465, opened 2022-11-04
Use run description for temporary env name if one is present

magni-:
Otherwise default to the existing behavior.

This will help clients with tons of concurrent temporary environments to distinguish between them:

image

### DIFF
--- a/rainforest/environments.go
+++ b/rainforest/environments.go
@@ -1,5 +1,7 @@
 package rainforest
 
+import "fmt"
+
 // EnvironmentParams are the parameters used to create a new Environment
 type EnvironmentParams struct {
 	Name        string `json:"name"`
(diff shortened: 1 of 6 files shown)


rainforest/environments.go
```diff
@@ -16,9 +18,14 @@ type Environment struct {
 
 // CreateTemporaryEnvironment creates a new temporary environment and returns the
 // Environment.
-func (c *Client) CreateTemporaryEnvironment(urlString string) (*Environment, error) {
+func (c *Client) CreateTemporaryEnvironment(runDescription string, urlString string) (*Environment, error) {
+	name := "temporary-env-for-custom-url-via-CLI"
+	if runDescription != "" {
+		name = fmt.Sprintf("%v-temporary-env", runDescription)
```
ubergeek42, line 24: Do we have any internal limits on the length of an environment name?

ubergeek42 (approved):
Looks ok to me, just a question about if we have any length limits on environment names.

rainforest/environments.go
magni-: We do, great catch 👍🏼

rainforest/environments.go
```diff
@@ -16,9 +18,17 @@ type Environment struct {
 
 // CreateTemporaryEnvironment creates a new temporary environment and returns the
 // Environment.
-func (c *Client) CreateTemporaryEnvironment(urlString string) (*Environment, error) {
+func (c *Client) CreateTemporaryEnvironment(runDescription string, urlString string) (*Environment, error) {
+	name := "temporary-env-for-custom-url-via-CLI"
+	if runDescription != "" {
+		if len(runDescription) > 241 {
```
pyromaniackeca, line 24: A bit late to the party, but probably worth using a constant instead of `241` to make it clearer what's going on.

PR #469, opened 2023-01-23
[RF-26747] Handle paginated responses from backend when fetching sites and environments

magni-:
Pagination code is copy-pasted from what we already do when fetching features and folders.

### DIFF
--- a/rainforest/resources.go
+++ b/rainforest/resources.go
@@ -190,36 +190,32 @@ type Site struct {
 
 // GetSites fetches sites available to use during the RF runs.
 func (c *Client) GetSites() ([]Site, error) {
-	// Prepare request
-	req, err := c.NewRequest("GET", "sites", nil)
-	if err != nil {
-		return nil, err
+	var sites []Site
+
+	collect := func(coll interface{}) {
+		newSites := coll.(*[]Site)
+		for _, site := range *newSites {
+			sites = append(sites, site)
+		}
 	}
 
-	// Send request and process response
-	var sitesResp []Site
-	_, err = c.Do(req, &sitesResp)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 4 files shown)


rainforest/resources.go
```diff
@@ -36,7 +36,7 @@ func (c *Client) getPaginatedResource(endpoint string, coll interface{}, collect
 
 	totalPagesHeader := res.Header.Get("X-Total-Pages")
 	if totalPagesHeader == "" {
-		return fmt.Errorf("Missing X-Total-Pages header in HTTP Response")
+		totalPagesHeader = "1"
```
ubergeek42, line 39: Do the newly paginated endpoints not have a total pages header? Or why this change? It seems fine, just curious.

jbarber, line 39: Is this because /sites + /environments are not paginated yet?

jbarber (approved):
Tests, or already has good coverage?

ubergeek42 (approved):
Looks ok to me

rainforest/resources.go
magni-: Exactly, this is to make the CLI forward compatible before adding pagination to those endpoints.

magni-: See response to @jbarber above, they're not paginated _yet_. Once they are, they'll have pagination headers like all the other paginated endpoints.

magni-:
> Tests, or already has good coverage?

There's already tests for `GetEnvironments` and `GetSites`, so they cover that switching to use `getPaginatedResource` doesn't break when the endpoints don't respond with an `X-Total-Pages` header. Actual pagination is tested in the resources that are already paginated (`GetFeatures, `GetFolders, `GetRunGroups`).

PR #471, opened 2023-01-25
[RF-26790] Update getBranchID to handle receiving multiple branches from the backend

magni-:
`GET /branches?name=<name>`'s behavior is changing to be a (case-insensitive, partial match) filter rather than a (exact-match) lookup. This PR updates `getBranchID` to handle that change by returning the branch with the expected name rather than blindly returning the first branch.


pyromaniackeca (approved):
👍

PR #472, opened 2023-01-26
[RF-26805] Handle paginated response from `GET /generators`

magni-:
I missed this in #469. We fetch generators when attempting to create a new one (to know whether we should fail or overwrite, depending on whether the `--overwrite` flag is used).

Since the backend is also adding filtering by type and name to the endpoint, I added those parameters as well in order to reduce the size of the response from the backend as much as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Rainforest CLI Changelog
-## Unreleased
+## 3.4.2 - 2023-01-26
+- Narrow down query when fetching fetching generators (tabular variables)
+  - (4df4734009a8c7a461bb8e0f6ee92082d91a4451, @magni-)
+- Handle paginated responses from backend when fetching generators (tabular variables)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
...
+## 3.4.2 - 2023-01-26
+- Narrow down query when fetching fetching generators (tabular variables)
+  - (4df4734009a8c7a461bb8e0f6ee92082d91a4451, @magni-)
+- Handle paginated responses from backend when fetching generators (tabular variables)
+  - (668e5245cd8884e5e1322499d0843c5388ba87d9, @magni-)
 - Handle multiple matching branches when merging or deleting a branch
   - (8c16acc9dcffdbf7a3cff509bcef143e7d915cc3, @magni-)
 
-## 3.4.1 - 2021-01-23
+## 3.4.1 - 2023-01-23
```
pyromaniackeca, line 10: 😄

pyromaniackeca (approved):
Looks good. 👍

PR #474, opened 2023-02-14
[RF-26554] Fix uploading a batch of new tests

magni-:
The backend API now requires `type` to be sent when creating a test. We added test types to the CLI in 832898b12ec58947b9710bd5d6e1231040646918, but only when uploading a single test. This wasn't caught in specs because we had no tests for uploading new tests, only for updating existing tests. I've updated the tests to cover:
- uploading a single new test
- uploading a single existing test
- uploading multiple tests (both new and existing)

The cleaner fix would be to DRY up the code between uploading a single RFML file or multiple files, but for now this is hopefully good enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Rainforest CLI Changelog
+## 3.4.3 - 2023-02-14
+- Fix `rainforest upload` command when uploading multiple tests including at least one new test
+  - (888fe5b10cfb4e7c9acd7164fbf24a41b6c755be, @magni-)
 ## 3.4.2 - 2023-01-26
 - Narrow down query when fetching fetching generators (tabular variables)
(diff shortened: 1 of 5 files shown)


pyromaniackeca (approved):
Not exactly sure what the test failure is about. but it seems to be test setup? Is it trying to write the files in parallel and complaining about that?

The code seems reasonable though, so approving with the assumption that the issue is with the test setup.

magni-:
> Not exactly sure what the test failure is about. but it seems to be test setup? Is it trying to write the files in parallel and complaining about that?
> 
> The code seems reasonable though, so approving with the assumption that the issue is with the test setup.

An issue with the test setup, which I've fixed using a mutex [per the docs](https://go.dev/doc/articles/race_detector#Typical_Data_Races)

PR #476, opened 2023-03-16
RF-27499 adding slimmer=true for tests endpoint since tests download is timing out for some clients

prabeshshrestha:
the CLI currently calls `tests` endpoint without slimmer=true, but it seems like things should just work with the slimmed down response. 
Because the response is timing out, some clients are getting `502` errors through the CLI.


magni- (commented):
Can you bump the version (patch to `3.4.4` is fine) and add a Changelog entry? 🙇🏼

magni- (approved):

PR #479, opened 2023-05-02
Allow specifying webhook when using custom-url

sebaherrera07:


jbarber (commented):
Have you tested this against the server?

jbarber:
> > Have you tested this against the server?
> 
> Not sure how to test it with the local code, is there any document on how to do that? I couldn't find it in the README.

Looks like you can configure the target server via the `RAINFOREST_API_URL` env var  and set this up to point at your local environment - but I don't think there's a problem with running against production using a single simple test?

jbarber:
> > Looks like you can configure the target server via the `RAINFOREST_API_URL` env var and set this up to point at your local environment - but I don't think there's a problem with running against production using a single simple test?
> 
> Actually I meant that I don't know how to run the local version of the cli itself, rather than running the published version. I'm OK with testing against production server. I guess there is a way to do that without publishing a new version of the cli?

Yep!
```
$ go build
$ ./rainforest-cli ...
```

jbarber (approved):
GTG as long as the data in the BE looks good with and without `--webhook`

Issue #473, opened 2023-02-08
--custom-url adds a trailing slash to URL


alanb43:
When using the "--custom-url" option in a github action (and possibly all of the time), the url that is passed to the test runner has a trailing slash. In my use case, the url I'm passing is a download link, and the slash is causing the download to not work / the response to be an XML error. Let me know if this is on my end!

magni-:
Hi @alanb43

The temporary environment's URL is properly created without a trailing slash, so there's no issue on the CLI (of GitHub Action)'s side of things. What's happening is that the test's `start_uri` is currently set to `/`, so the test starts off navigating to `{environment_url}/`.

If you set your tests' `start_uri` to the empty string, your tests will start off navigating to `{environment_url}`. This must currently be done via the API, see [the `put-test` endpoint](https://help.rainforestqa.com/reference/put-test).

Alternatively, you can add a Navigate step, navigating to `{{ rainforest.url_base }}`. This will navigate to `{environment_url}`, regardless of what the test's `start_uri` is.

magni-:
A hackier solution would be to append `#` or `?` to the `--custom-url`. This means the file will be properly downloaded even with an extra `/` at the end of the URL.

PR #487, opened 2023-08-15
Allow creation of temporary environment using custom-url with empty fragment

philip-maina:
## Summary
Prior to this PR, if you passed a `custom-url` with an empty fragment e.g `https://example.com#` the `#` would be stripped off and a temporary environment with url `https://example.com` would be created instead. 

This happens because we parse the `custom-url` (see [here](https://github.com/rainforestapp/rainforest-cli/blob/d6624ff90357a64180d43e525c24650713544018/runner.go#L420)) and parsing urls like so `url.Parse(myUrl)` in golang has a [known issue](https://github.com/golang/go/issues/29603). 

The fix implemented ensures that `#` is retained in the url even if the fragment is empty.

Related issue: #473

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -437,7 +437,7 @@ func (r *runner) makeRunParams(c cliContext, localTests []*rainforest.RFTest, br
 		}
 
 		var environment *rainforest.Environment
-		environment, err = r.client.CreateTemporaryEnvironment(description, customURL.String(), webhookURL.String())
+		environment, err = r.client.CreateTemporaryEnvironment(description, customURLParam, webhookParam)
 		if err != nil {
 			return rainforest.RunParams{}, err
 		}
--- a/runner_test.go
+++ b/runner_test.go
@@ -82,6 +82,9 @@ type fakeRunnerClient struct {
 	mu sync.Mutex
 	// "inherit" from RF API
 	testRfAPI
+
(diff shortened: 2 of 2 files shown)


sebaherrera07 (commented):
Looks good to me. Did you try this change yourself locally?
Just to be sure that it works as expected.

runner.go
```diff
@@ -437,7 +437,7 @@ func (r *runner) makeRunParams(c cliContext, localTests []*rainforest.RFTest, br
 		}
 
 		var environment *rainforest.Environment
-		environment, err = r.client.CreateTemporaryEnvironment(description, customURL.String(), webhookURL.String())
+		environment, err = r.client.CreateTemporaryEnvironment(description, customURLParam, webhookURL.String())
```
magni-, line 440: Should we also use `webhookParam` here, since that could be susceptible to the same issue?

magni- (approved):

philip-maina:
> Looks good to me. Did you try this change yourself locally?
Just to be sure that it works as expected.

Here's the run locally 

Screenshot 2023-08-16 at 17 40 02
Screenshot 2023-08-16 at 17 40 35

magni-:
@philip-maina Worth a `3.5.1` release?

PR #490, opened 2024-06-20
RF-34736 - Rainforest Direct Connect

ubergeek42:
This adds the `direct-connect` subcommand for establishing a Rainforest Direct Connect session, enabling the Rainforest test environments access to the network where this command is run.

It's intended to be used like `rainforest-cli direct-connect` or `rainforest-cli direct-connect --tunnel-id c12345`

Upgrading go was required for the wiretap dependency, but also good to put everything in sync (previously go.mod specified 1.12, while we were building with 1.17). Now everything is 1.22.

I don't have any integration test yet, because the backend doesn't actually support Direct Connect for runs (only for previews right now). I think it's something we'll have to come back to., and I created a ticket to track that [RF-34749](https://rainforest.atlassian.net/browse/RF-34749)

[RF-34749]: https://rainforest.atlassian.net/browse/RF-34749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


direct_connect.go
```diff
@@ -0,0 +1,187 @@
+package main
```
ubergeek42, line 1: I didn't add tests for this yet, but I think the best/correct thing to do is an integration test, because there's no good way to do some unit test for it.

direct_connect.go
```diff
@@ -0,0 +1,187 @@
...
+	"wiretap/peer"
+	"wiretap/transport/icmp"
+	"wiretap/transport/tcp"
+	"wiretap/transport/udp"
+
+	"github.com/rainforestapp/rainforest-cli/rainforest"
+	"github.com/urfave/cli"
+)
+
+func launchDirectConnect(c cliContext, rfApi *rainforest.Client) error {
```
ubergeek42, line 31: much of this function is a trimmed down version of what's in wiretap's serve cmd.

rainforest-cli.go
```diff
@@ -635,8 +635,22 @@ func main() {
...
 		},
+		{
+			Name:         "direct-connect",
+			Usage:        "Start a Rainforest Direct Connect session",
+			OnUsageError: onCommandUsageErrorHandler("direct-connect"),
+			Description:  "Starts a Direct Connect Session for the given tunnel-id",
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "tunnel-id",
+					Usage: "tunnel-id",
```
jbarber, line 646:
```suggestion
					Usage: "tunnel-id",
					Value: "default",
```

Let's you remove  L33-35 in `direct_connect.go`?

rainforest/direct_connect.go
```diff
@@ -0,0 +1,36 @@
+package rainforest
+
+// DirectConnectParams are the parameters the Rainforest Direct Connect api expects
```
jbarber, line 3:
```suggestion
// DirectConnectParams are the parameters the Rainforest Direct Connect API expects
```

rainforest/direct_connect.go
```diff
@@ -0,0 +1,36 @@
...
+type DirectConnectParams struct {
+	PublicKey string `json:"public_key"`
+	ID        string `json:"id"`
+}
+
+// DirectConnectConnection is the response from Rainforest's Direct Connect api telling us how to connect
```
jbarber, line 9:
```suggestion
// DirectConnectConnection is the response from Rainforest's Direct Connect API telling us how to setup the wireguard tunnel
```

rainforest/direct_connect.go
```diff
@@ -0,0 +1,36 @@
...
+type DirectConnectConnection struct {
+	ID              string `json:"id"`
+	ServerPublicKey string `json:"server_public_key"`
+	ServerPort      int    `json:"server_port"`
+	ServerEndpoint  string `json:"server_endpoint"`
+}
+
+// ConfigureDirectConnect tells Rainforest to set up a direct connect tunnel for us
```
jbarber, line 17:
```suggestion
// ConfigureDirectConnect tells Rainforest to set up a Direct Connect tunnel for us
```

Call the function `SetupDirectConnectTunnel` then?

direct_connect.go
```diff
@@ -0,0 +1,187 @@
...
+		log.Panic(err)
+	}
+
+	http.HandleFunc("/ping", handlePing)
+
+	err = http.Serve(listener, nil)
+	if err != nil {
+		log.Panic(err)
+	}
+}
```
jbarber, line 178: Agree it's not worth unit testing the rest, but add unit tests for `apiHandler` / `handlePing`?

jbarber (approved):

direct_connect.go
```diff
@@ -0,0 +1,187 @@
...
+	var (
+		wg   sync.WaitGroup
+		lock sync.Mutex
+	)
+
+	// Static configuration settings
+	listenPort := 51820
+	persistentKeepaliveInterval := 25
+	mtu := 1420
+	allowedIPs := []string{"198.18.18.1/32", "fd:16::1/128"}
```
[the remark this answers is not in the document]
ubergeek42: Because this is purely a userspace implementation of wireguard, I think there is no conflict if the host has the same IPs in use. It just means you won't be able to access the host directly over the tunnel, because it'll be shadowed by the wireguard userspace tunnel. I think this isn't much of a concern.

And yes, that was more or less my logic re: IPv6; I figured there wasn't any harm leaving it here in case we support it in the future it's one less place to change.

direct_connect.go
```diff
@@ -0,0 +1,187 @@
...
+		wg.Done()
+	}()
+
+	// Start API handler.
+	wg.Add(1)
+	go func() {
+		apiHandler(tnetWireguard, ipv4Addr, uint16(80))
+		wg.Done()
+	}()
+	log.Println("Rainforest Direct Connect running! Press Ctrl+C to exit")
```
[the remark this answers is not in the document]
ubergeek42: `listenPort` is fixed at 51820 (L58 above), so you can't run multiple instances on a single host currently. I would also expect that if you were running multiple, you'd probably throw them into a simple systemd service and then not care about PIDs or anything like that, letting systemd handle it. But it should be easy enough to add, and could prove useful in some circumstances.

direct_connect.go
ubergeek42: `apiHandler` is likewise difficult to test, because it needs a netstack.Net and binds/listens on a network interface (which is annoying to mock). But handlePing is simple enough, so I'll add that.

jbarber: Fair enough. Could avoid mocking and just grab a random IPv4/6 link [local address](https://en.wikipedia.org/wiki/Link-local_address)? Although I guess some hosts will not have these and would cause the tests to fail spuriously.

PR #495, opened 2024-06-27
RF-34749 - Add integration test for Rainforest Direct Connect

ubergeek42:
This launches direct-connect in the background, along with a simple python http server, and then kicks off a rainforest run that should connect to the python server and verify the index.html says RainforestCLI Direct Connect Test.

The account this is configured to use has a direct connect tunnel (id 6), and the run group is configured to use that tunnel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -260,6 +260,29 @@ jobs:
       - store_artifacts:
           path: dist/checksums.txt
 
+  direct_connect_integration_test:
+    executor: linux
+    steps:
+      - checkout
+      - run: git submodule update --init
+      - go_build
+      - run:
+          name: Start the direct connect tunnel
+          command: |
+            RAINFOREST_API_TOKEN=$RAINFOREST_DC_API_TOKEN ./rainforest --skip-update direct-connect --tunnel-id 6
+          background: true
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -260,13 +260,44 @@ jobs:
...
+  direct_connect_integration_test:
+    executor: linux
+    steps:
+      - checkout
+      - run: git submodule update --init
+      - go_build
+      - run:
+          name: Start the direct connect tunnel
+          command: |
+            RAINFOREST_API_TOKEN=$RAINFOREST_DC_API_TOKEN ./rainforest --skip-update direct-connect --tunnel-id 6
```
sebaherrera07, line 272: I think the `tunnel-id` argument is optional, you can remove it if you want.

sebaherrera07 (approved):
🚀

.circleci/config.yml
ubergeek42: I wanted to leave it to be explicit, that way if someone tries to swap a token things don't break in weird ways, instead the tunnel will fail to start properly with some authentication error.

sebaherrera07: Makes sense!

PR #496, opened 2024-07-30
RF-35226 remove default tunnel id argument value

sebaherrera07:
* Remove the `"default"` value for the `tunnel-id` argument for direct connect.
* Output the connection name to console when the tunnel is established.
* Add instruction to README to install submodules.

image


ubergeek42 (approved):
Thanks for updating the readme!

